PR #4, opened 2020-05-14
Example structure.

timsifive:
This combines most of what we talked about in this week's meeting.
Definitely need more details about Physical Memory (is that even a task
group), and caches.

### DIFF
--- a/example.md
+++ b/example.md
@@ -63,5 +63,7 @@ description, etc.
     * tuple of LOW, HIGH addresses
         * cacheable
         * LR/SC support
+        * alignment and size restrictions
+        * mode restrictions
     * tuple of LOW, HIGH addresses
         * ...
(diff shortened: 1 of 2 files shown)


allenjbaum:
Looking through your example:
\item triple of VALUE0, MASK0
That's a tuple, not a triple

timsifive:
Thanks. There are a few typos in this file. However, this file does not include "\item" anywhere, which makes me wonder if you were looking at this example.md, or at something we previously wrote in LaTeX.

allenjbaum:
I looked at the link, couldn’t find pre-built .pdf, looked at the changes, and that’s what popped up. It may have been example.md, I’ll go back and look.

-Allen

timsifive:
Merging, to make it easier to access these examples. Let's get all our examples here so they're in one place.

PR #5, opened 2020-07-08
Add charter in README

timsifive:


README.md
```diff
@@ -0,0 +1,28 @@
...
+The configuration structure should:
+* be flexible enough that future task groups won’t feel the need to
+  create another structure used to describe implementation parameters.
+* be relatively easy to translate into Devicetree
+
+The configuration structure is intended to be used:
+* to describe RISC-V hardware profiles
+* by firmware and BIOSes during the boot process
+* by debuggers
+* by a tool chain to build software tailored to a configuration
```
JohnAZoidberg:
```suggestion
* by a tool chain to build software tailored to a configuration profile
```

Since this is supposed to be about profiles.

changab (approved):
the change looks good to me

changab (approved):
Much clear now

JohnAZoidberg (approved):

simon5656:
I would have thought a key consumer of the configuration data are the simulators used in the compliance process - and also for design verification - so I think the charter should also include the use of the config data to configure reference simulators. Are you aware of the https://github.com/riscv/riscv-config work to configure simulators to reflect device options we have been doing in the compliance working group?

timsifive:
If you want to expand the charter, please discuss that on the mailing list, which gets a lot more eyes than github.

timsifive:
Not really relevant anymore.

PR #8, opened 2020-07-21
Add simple CBOR example

JohnAZoidberg:
Here's a very simple example of a binary format with CBOR and CDDL.
There's a CDDL schema and a JSON input file.

As you can see, the example is quite big (667 bytes) but it contains mostly the strings, which I'm planning to get rid of.

I looked at the CBOR spec in detail and it can be quite compact if we're smart about it. It seems that an optimized CBOR implementation could be as small as a custom format. But the advantage of CBOR is, that parsers already exist for many languages and requirements.

###### Files

- `spec.cddl` defines the format of the CBOR
- `instance.json` is an example instance, just like `../example.md`
- `instance.cbor` is the example converted to CBOR
- `test.sh`
  - Generates a random JSON instance, in compliance with the spec
  - Validates the example instance
  - Converts the example instance to CBOR
  - Shows the converted CBOR as hex 

###### TODO
- [ ] Eliminate string keys from CBOR
- [ ] Represent e.g. bitfields more compactly
- [ ] Use something nicer than JSON as the input format (e.g. Dhall, JSON5, Jsonnet, ...)


timsifive:
That CDDL file is really interesting. Thanks for taking this on. This looks very promising.

JohnAZoidberg:
Contained in #9

PR #10, opened 2020-07-22
Expand Protobuf example

JohnAZoidberg:
Here's a protobuf version with all of the config, like the Markdown, Scala, Dhall and CBOR examples.

Binary size is 196 bytes.

###### TODO
- [ ] Choose a good input format (JSON is possible but I haven't found out-of-the-box tools)
- [ ] Optimize e.g. bitfields
- [ ] Make sure the python script works with the new schema


JohnAZoidberg:
> One question: Is it possible for a parser that doesn't know the full CDDL to ignore parts of the binary format that it doesn't know about?

Answering Tim's question about my CBOR example in relation to Protobuf:
The parser can ignore unknown fields, unlike CBOR you *cannot* tell anything about the structure of the unknown fields. In CBOR you could parse them and pretty print them to the user.

For example: you as a parser encounter a field named `debug` that you don't know exists. The writer of the config structure renamed the `debugging` field to `debug` and put exactly the same information there. With Protobuf you cannot read it, with CBOR you can.
I realize this is a scenario we can try to avoid, so I'm not sure if it's a big drawback of protobuf.

PR #11, opened 2020-07-24
Dhall example

JohnAZoidberg:
Based on #12

###### TODO

- [ ] Add README explaining the Dhall examples and all of the files


JohnAZoidberg:
Let's merge this also, already then I can easily create a Dhall to CBOR example next week.

PR #13, opened 2021-02-03
Read schema with file handle opened with "Read" and "Binary" flags.

changab:


timsifive (approved):
Oops. Thanks for fixing that.

PR #14, opened 2021-02-03
Skip unsupported configuration structure type

changab:


binary/custom-encoding/cs_decode/cs_decode.c
```diff
@@ -106,7 +106,7 @@ static bool needs_length(cs_decoder_t *decoder, unsigned type)
 static int decode_schema_type(cs_decoder_t *decoder, unsigned type)
 {
     debug(decoder, "decode_schema_type(%d)\n", type);
-
+    unsigned i;
```
timsifive: Can you move this inside `while (decoder->offset < end)`? That way the for loop on line 120 keeps the variable local, and somebody reading the code doesn't have to worry about whether `i` there is used later.

timsifive (changes requested):
Looks good, I just have a style comment below.

binary/custom-encoding/cs_decode/cs_decode.c
changab: Done. but I just force pushed to upstream by accident. Should force push to my personal repo and have a PR for it.

timsifive: I'm not sure exactly what you did. master in this repo seems to be untouched, and this PR was updated correctly. So whatever it was it worked out fine. :-)

timsifive (approved):
Looks good, thanks!

PR #19, opened 2021-02-08
Decrease decoder size

timsifive:
Net change in .o files:
Old .o totals: 2646, 2978
New .o size: 2362, 2458

There's some more low-hanging fruit here.


changab (approved):
I didn't run the encoder and decoder with these patches. However, the code logic looks good to me.

PR #20, opened 2021-02-19
Use smaller data types in cs_decode.

timsifive:
Old total .o size: 2354, 2450
New total .o size: 1596, 1678

### DIFF
--- a/binary/custom-encoding/cs_decode/cs_decode.h
+++ b/binary/custom-encoding/cs_decode/cs_decode.h
@@ -11,9 +11,19 @@ enum {
     BUILTIN_END
 };
 
+/*
+ * There are a bunch of uint8_t below, to save space in the binary. With larger
+ * schemas that will have to be uint16_t. (Maybe even uint32_t for some huge
+ * schemas.)
+ * The goal is to have tool.py generate this entire file, and generate the
+ * appropriate types based on the schema itself. Until that is done, I'm just
+ * changing the types to be the smallest. The compiler will warn/error when
+ * generated constants don't fit in these types.
+ */
+
 typedef struct {
-    unsigned code;
-    unsigned type;
+    uint8_t code;
+    uint8_t type;
 } cs_typedef_entry_t;
 
 #define CS_FLAG_REPEATABLE      1
@@ -21,12 +31,12 @@ typedef struct {
 typedef uint8_t cs_typedef_flags_t;
 
 typedef struct {
-    unsigned entry_count;
-    unsigned entry_index;
+    uint8_t entry_count;
+    uint8_t entry_index;
 } cs_typedef_t;
 
 typedef struct {
-    unsigned type_count;
+    uint8_t type_count;
     const cs_typedef_t *types;
     const cs_typedef_entry_t *all_entries;
     const cs_typedef_flags_t *all_flags;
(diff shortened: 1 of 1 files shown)


binary/custom-encoding/cs_decode/cs_decode.h
```diff
@@ -11,30 +11,40 @@ enum {
...
 
 #define CS_FLAG_REPEATABLE      1
 #define CS_FLAG_REQUIRED        2
 typedef uint8_t cs_typedef_flags_t;
 
 typedef struct {
-    unsigned entry_count;
-    unsigned entry_index;
+    uint8_t entry_count;
+    uint8_t entry_index;
```
changab, line 35: Shall we leave this still unsigned? or at least the unsigned __int16. entry_index seems to be very easy to go over 256.
e.g. debug_trigger_mcontrol already occupied 34 entries.

timsifive: My longer term goal is to make tool.py generate this file, so it can pick the smallest type that fits the numbers actually used by the schema.

This change just lets us see what the actual size improvement is when I implement that.

This is not a permanent limit to 256 entries. I agree that there will be cases where we'll go over that. (But it's probably not common, especially for smaller schemas in use by smaller firmware, so the size optimization is worth it.)

changab: Do you mean the datatype of entry_index in this structure would be different according to how many entries are used?

timsifive: Yes. As tool.py processes the schema and creates schema.c, it knows the max value required for entry_count and entry_index. If it knows the max value for a type is <= 255 it can use uint8_t and so on. Then it creates cs_decode.h (or maybe that's moved somewhere else) using the appropriate types.

changab: This way is ok for the standalone parser as the application or OEMs clone parser from Github and change it to whatever they want for their firmware code. But to have tool.py to generate the different data types in the header file is the problem to open source firmware solution. We would only have one parser in the firmware code base, that parser should be able to deal with either small or large CS binary. But the header file shouldn't be changed in the core code or require customers to rebuild the firmware if they get the new CS binary (which turns small data type into large data type) from the vendor.
The alternative could be, in the opensource firmware, we can define two structures in the header file to refer to the small datatype or the big datatype, but this requires a bit as the header at the beginning of CS binary which indicates the binary has 8-bit or 16-bit index.

timsifive: The data types do not depend on the CS binary. They only depend on the schema. This change doesn't affect any functionality, just the size of the code.

changab: ok. I got the point.
So in the edk2 or opensbi, we can just use the bigger data type (e.g. int16) for the entry_index in the parser code.

timsifive: Yes, if the schema you're using has too many entries for uint8 to be sufficient, you'll use uint16.

timsifive: Assuming it all makes sense now, can you approve the pull request? It's mostly a formality, but it's a good habit to get into.

changab (approved):
All good to me now.

PR #24, opened 2021-03-13
Add custom support for fixed-width numbers.

timsifive:
Supports any number of bits from 1 to 128. The user can use types by specifying Int1, Int2, Int3, all the way up to Int128.

To keep this efficient, I pulled the separate cs_decode.[ch] into the python tool, which now reads c_code.c and c_header.h and puts the contents into the files it generates.

Code size did go up a bit, from 1774/1856 to 1819/1873.

### DIFF
--- a/binary/custom-encoding/c_code.c
+++ b/binary/custom-encoding/c_code.c
@@ -2,8 +2,6 @@
 #include <stdio.h>
 #include <stdarg.h>
 
-#include "cs_decode.h"
-
 typedef struct {
     const cs_schema_t *schema;
     const uint8_t *encoded;
@@ -12,6 +10,10 @@ typedef struct {
     unsigned offset;
     cs_path_t path;
 } cs_decoder_t;
+
+static inline bool is_builtin(unsigned type);
+static int decode_builtin(cs_decoder_t *decoder, unsigned type);
+static int needs_length_builtin(unsigned type);
 
 static void debug(const cs_decoder_t *decoder, const char *format, ...)
 {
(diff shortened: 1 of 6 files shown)


changab:
The idea of this change is good. However the concerns of some tool-generated code for the firmware are,
- Some static functions in schema.c are better to be public functions, such as needs_length_builtin(), decode_builtin(), is_builtin. Maybe as well as decode_schema_type and some others that I didn't check in deep. the reason is the OEM code for their proprietary CS can leverage the above functions. 
- Because the BUILDIN_NUMBER and all possible BUILTIN_FIXEDn are primitives in CS, we should generate code for all of those in the function like decode_builtin() and other similar ones. So OEM can leverage the function without the problem if they use BUILTIN_FIXED3 which doesn't have the corresponding code in schema.c.
- We should define all possible BUILTIN_FIXEDn in the schema, so OEM doesn't have to define it.

I understand the implementation you have now is small and the code is generated on demand. In order to have a common CS decoder on the firmware solutions, the above should be considered.
But I think don't have to do it right away, maybe just wait until we choose custom binary as the format of CS.

changab (approved):

timsifive:
It sounds like in the future we want two source options:
1. Make the smallest possible parser for this schema.
2. Make a parser that has code that can be useful to parse structures that are not part of this schema.

I think the overall code might be smaller if we generate 2 schema-specific parsers than if we have a single generic parser with shared code. We'll have to experiment with that, though.

Does that sound right? I must admit I'm still not entirely clear on the OEM use case you have in mind. Let's talk about that in email.

PR #25, opened 2021-04-19
The value of Configuration-Structure-Schema ASN.1

changab:
- The size of UPER binary 100 bytes
- Some minor changes on schema.asn.

### DIFF
--- a/binary/asn1/schema.asn
+++ b/binary/asn1/schema.asn
@@ -1,16 +1,16 @@
 Configuration-Structure-Schema DEFINITIONS AUTOMATIC TAGS ::= BEGIN
    Range ::= SEQUENCE {
       start INTEGER,
-      length INTEGER
+      length INTEGER 
    }
    FlexibleRange ::= SEQUENCE {
-      single SEQUENCE OF INTEGER OPTIONAL,
-      start SEQUENCE OF INTEGER OPTIONAL,
-      end SEQUENCE OF INTEGER OPTIONAL,
+      single INTEGER OPTIONAL,
+      start INTEGER OPTIONAL,
+      end INTEGER OPTIONAL,
       ...
    }
    Hart ::= SEQUENCE {
-      hartid FlexibleRange,
+      hartid SEQUENCE OF FlexibleRange,
       debug Debug OPTIONAL,
       isa Isa OPTIONAL,
       privileged Privileged OPTIONAL,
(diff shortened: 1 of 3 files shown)


binary/asn1/schema.asn
```diff
@@ -33,7 +33,7 @@ Configuration-Structure-Schema DEFINITIONS AUTOMATIC TAGS ::= BEGIN
       ...
    }
    Configuration ::= SEQUENCE {
-      harts SEQUENCE OF Hart OPTIONAL,
+      harts SEQUENCE (SIZE(1..64)) OF Hart OPTIONAL,
```
timsifive, line 36: I don't think we can limit the size like that. E.g. Esperanto has a core with more than 1000 harts already.

binary/asn1/schema.asn.value
```diff
@@ -0,0 +1,203 @@
...
+                  matchEqual TRUE,
+                  matchNapot TRUE,
+                  matchGreaterEqual TRUE,
+                  matchLess TRUE,
+                  matchLowMask TRUE,
+                  matchHighMask TRUE,
+                  matchNotEqual TRUE,
+                  matchNotNapot TRUE,
+                  matchNotLowMask TRUE,
+                  matchNotHighMask TRUE,
```
timsifive, line 38: In the other example these are false, and some others are as well.

binary/asn1/schema.asn.value
```diff
@@ -0,0 +1,203 @@
...
+                  timingBefore FALSE,
+                  timingAfter FALSE,
+                  sizeAny FALSE,
+                  sizeS8 FALSE,
+                  sizeS16 FALSE,
+                  sizeS32 FALSE,
+                  sizeS64 FALSE,
+                  sizeS80 FALSE,
+                  sizeS96 FALSE,
+                  sizeS112 FALSE,
```
timsifive, line 65: In this format is it allowed to omit default (false) values? That would be really helpful for users writing descriptions.

binary/asn1/schema.asn
changab: This is just the experiment of using SIZE syntax to see if we would get the reduced size. The entire schema should still have room to tailor for the concise size.

binary/asn1/schema.asn.value
changab: I will check it over again.

binary/asn1/schema.asn.value
changab: Hmm, I remember there is something like Default syntax, not sure how to use it and what is the advantage. I will check it.

PR #28, opened 2021-05-22
Use SEQUENCE of CHOICE types instead of OPTIONAL.

timsifive:
We're concerned that we'll end up with a ton of hart extensions, and
each optional one will use a bit even if not mentioned at all. A
sequence of choices scales better.

With our current example, the binary size increases from 110B to 116B.
However, if I add 100 dummy extensions, the OPTIONAL version becomes
172B while the SEQUENCE of CHOICE version only increases to 119B. I
think it is worth getting this scalability. We might want to use the
same structure in a few other places as well.

There is another problem. asn1tools refuses to use CHOICE with trees
that come out of python's yaml loader. The problem is that it wants a
CHOICE to be represented as a tuple, and the yaml loader turns them into
a list. To work around this I converted the yaml to jer. jer is more
constrained and doesn't allow comments, but for now at least it works.
Long term we should figure out a way to parse something friendlier.

examples/debug_module.jer is 20B in UPER
examples/example.jer is 116B in UPER

Note: This commit builds on #27, so that should be reviewed/merged
first.


changab:
> With our current example, the binary size increases from 110B to 116B.

Each OPTIONAL item requires one bit in the preamble. Each CHOICE requires an Index in the preamble to indicate which type is chosen, and the number of bits for the Index of CHOICE is the minimum bits for the types to choose.
Says we have 10 optional items, then 10 bits in the preamble indicates if the item exists.
That uses 4-bit of the index for CHOICE, 4bits x 10 (40 bits) if all items are chosen. This is my understanding of CHOICE, am I wrong?

> However, if I add 100 dummy extensions, the OPTIONAL version becomes
> 172B while the SEQUENCE of CHOICE version only increases to 119B. I
> think it is worth getting this scalability. We might want to use the
> same structure in a few other places as well.


Do you choose all 100 dummy extensions?

I also did some experiments of ASN.1 optimization, which can share in next week's meeting.

timsifive:
Your explanation matches my understanding.

In the numbers I reported I did not provide any data for the dummy extensions, so they were just providing more unused possibilities.

changab (approved):
Didn't have a closer look at those values. Code and schema look good to me.

PR #29, opened 2021-05-26
Optimize schema.asn

changab:
Optimize schema.asn in below manners,
- Define different sizes of integer
- Define different sizes of Range
- Define different sizes of FlexibleRange
- Limit the number of bits used for the count of "SEQUENCE OF single"
- Define variant of hartid based on different sizes of integer.

examples/example.asn is 91B in UPER

Total encoded length = 90.6
Encoded successfully in 91 bytes:
702A2001 04091010 10001030 10104FFF FFFFFA40 20208020 20800000 00052026
00820290 08082008 08200000 0001520A 010B64FD F44B8109 49D02A40 4C014604
24680424 68026802 92020420 00023E40 40400041 1C040004 000480

### DIFF
--- a/examples/example.jer
+++ b/examples/example.jer
@@ -2,21 +2,23 @@
     "harts": [
         {
             "hartid": {
-                "single": [
-                    0
-                ]
+                "hartId4": {
+                    "single": [
+                        0
+                    ]
+                }
             },
             "extension": [
                 {
                     "debug": {
                         "trigger": [
                             {
-                                "index": {
-                                    "start": [
-                                        0
-                                    ],
-                                    "end": [
-                                        3
+                                "index" : {
+                                    "range": [
+                                        {
+                                            "start": 0,
+                                            "length": 3
+                                         }
                                     ]
                                 },
                                 "mcontrol": [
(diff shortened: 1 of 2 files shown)


timsifive (commented):
Can you update example.jer instead of creating a new example.asn? That way everything can be converted/tested with rvcs.py. asn1tools doesn't seem to have support for this file format, otherwise I'd simply add it.

changab:
Hmm, ok. so no ASN.1 syntax example? JSON only?

timsifive:
Yes, please.

I'm not really a fan of the JSON format, but it's the only thing that works with asn1tools, which in turn is the only real tool we have that we can recommend that people use. Having them generate configuration structures by going through some website isn't going to work.

I still plan to make a YAML parser for asn1tools so we can have comments, and some other convenience features. Once that's done I'll convert all the examples to YAML. Having examples in multiple formats is too confusing.

(I do like how this ASN.1 syntax handles choices, but I just don't know how to parse it.)

changab:
example.jer updated.

timsifive (approved):
Cool. This is a really interesting optimization.

PR #32, opened 2021-06-07
Remove ... from FlexibleRange types.

timsifive:
Surely we'll never need anything besides single and range?
Saves a byte frome example.jer.


changab (approved):
Looks good

PR #33, opened 2021-06-08
Fix rvcs test on Windows.

timsifive:
Don't use tempfile and then open that same file again. In fact, there's
no need for a temporary file at all.

Also mention how to run the test in README.md.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # RISC-V Configuration structure
 
-Layout of files here:
+## Layout of files here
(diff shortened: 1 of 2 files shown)


changab (approved):

timsifive:
Are you able to confirm that it now works on Windows?

changab:
Yes, I can run it successfully.

timsifive:
Thanks for testing. :-)

PR #35, opened 2021-06-17
Rename FooNN to use width instead of range.

timsifive:
Now Integer4 represents a 4-bit integer, which eliminates Integer16 used
to refer to a 4-bit integer. (My programmer brain sees Integer16 and
parses it as a 16-bit integer, and I assume other people will make the
same association.)

### DIFF
--- a/schema/configuration-structure.asn
+++ b/schema/configuration-structure.asn
@@ -3,7 +3,7 @@ DEFINITIONS
    AUTOMATIC TAGS ::=
 BEGIN
    IMPORTS Debug, DebugModule FROM Debug-Extension
-      Range, FlexibleRange, FlexibleRange4, FlexibleRange8, FlexibleRange16 FROM Helper-Types
+      Range, FlexibleRange, FlexibleRange2, FlexibleRange3, FlexibleRange4 FROM Helper-Types
       Tuple, Triple, PossibleValue, PhysicalAddress FROM Helper-Types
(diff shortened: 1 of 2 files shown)


schema.asn
```diff
@@ -34,27 +34,27 @@ Configuration-Structure-Schema DEFINITIONS AUTOMATIC TAGS ::= BEGIN
...
 
    -- 4-bit flexible range [0..15]
    FlexibleRange16 ::= SEQUENCE {
       -- The count of single sequence [1..16] which is
       -- a 4-bit count
-      single SEQUENCE SIZE(1..16) OF Integer8 OPTIONAL,
+      single SEQUENCE SIZE(1..16) OF Integer3 OPTIONAL,
       -- The intenger range is [0..15]
-      range SEQUENCE OF Range16 OPTIONAL
+      range SEQUENCE OF Range4 OPTIONAL
```
changab: also change FlexibleRangeN to width base? FlexibleRange2, 3, 4?

changab:
Not sure why this PR has 22 commits for different issues, some already reviewed as I can remember. Only one comment, all others look good.

changab (approved):

schema.asn
timsifive: Good point. Will do.

timsifive:
Hmm... yeah the PR is a mess when I resolved conflicts. Let me fix that and force push.

changab (approved):

PR #37, opened 2021-06-29
Add Custom types for vendor-specific extensions.

timsifive:
Identify vendors using JEDEC code. Use OCTET STRING for the undescribed
data.


schema.asn
```diff
@@ -61,6 +61,32 @@ Configuration-Structure-Schema DEFINITIONS AUTOMATIC TAGS ::= BEGIN
...
+      data OCTET STRING
+   }
+
+   Custom ::= SEQUENCE {
+      -- Bits 6:0 of the designer/manufacturer’s Identification Code as assigned
+      -- by JEDEC Standard JEP106.
+      jedecLow INTEGER (0..127),
+      -- The modulo 32 count of the number of continuation characters (0x7f) in
+      -- that same Identification Code.
+      jedecContinuation INTEGER (0..31),
```
changab: Suppose this ID  is the same as the mvendorid of the hart, right?
So firmware can use this ID to validate if the custom data exactly for this hart.

schema.asn
```diff
@@ -73,6 +99,7 @@ Configuration-Structure-Schema DEFINITIONS AUTOMATIC TAGS ::= BEGIN
       privileged Privileged OPTIONAL,
       clic Clic OPTIONAL,
       fastInt FastInt OPTIONAL,
+      custom SEQUENCE OF Custom OPTIONAL,
```
changab: If the Custom data only for this hart, then we don't have to use SEQUENCE OF for Custom, which results in the duplicate jedecLow and jedecContinuation in each Custom instance.
"entries SEQUENCE OF" can support multiple custom data for the same vendor on this hart.

schema.asn
timsifive: Good point. I should phrase this to use the same kind of definition/language as we have for mvendorid.

schema.asn
timsifive: I imagine a system could contain custom information from multiple vendors, so we do need a SEQUENCE OF Custom. But you're right that inside Custom we can include a SEQUENCE OF. I'll redo this a bit.

timsifive: I take this back. I think the schema already does what you want.

Custom contains a vendorId, and a SEQUENCE OF CustomOctetString. So for each vendor ID you can include a bunch of different OCTET STRINGs without repeating the vendor ID.

changab: My concern was the " SEQUENCE OF" in "**custom SEQUENCE OF Custom OPTIONAL**". this leads to an additional one byte for the number of Custom type instances. I thought there would be only one vendor of the custom type for this HART, however "SEQUENCE OF" is fine if multiple vendors of the custom type also the possible use case. Or we can have 

    custom SEQUENCE OF Custom OPTIONAL
    customSingle Custom OPTIONAL

This can save 7-bit (one for OPTIONAL) if only one vendor for the custom type if we care about that 7 bits. :)
Or use CHOICE to choose the single custom type or multiple custom types.

I am also ok with the current PR. What is your thought?

changab (approved):

PR #38, opened 2021-07-09
Crude conversion to asciidoc.

timsifive:
There's more to be done, but let's get off the LaTeX train before we've
written much of anything.


timsifive:
I used `pandoc` for the translation.

changab (approved):
I didn't go through the adoc carefully. I think the text is from what we had before and you just covert it to adoc format, right?
I agree with moving to adoc and we can start to revisit it based on adoc.

PR #39, opened 2021-07-15
Flesh out Debug section

timsifive:
Refactored the schema into multiple DEFINITIONS so that each extension can provide its own one. That avoids name space conflicts, and makes it clear who "owns" what part of the document.

Some small choices I made:
1. Using fooSupported to indicate whether foo is supported or not, while I'm just using `bar` if there is a bit called `bar` that always reads as one.
2. Allow full description of every implementation decision that software might care about. However, the stuff that can be trivially read from the hardware is put into the optional Complete section so that small descriptions don't have to include it.

Modified rvcs.py so that `test` can now deal with default values. (This was a problem before because it would constitute a value not specified in the input but showing up in the output.)

examples/debug_module.jer is 18B in UPER
examples/example.jer is 90B in UPER

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -1,9 +1,9 @@
 name: Schema/Example Test
 on: [push]
 jobs:
-    Explore-GitHub-Actions:
+    Test-Examples:
         runs-on: ubuntu-latest
         steps:
             - uses: actions/checkout@v2
-            - run: pip install asn1tools
+            - run: pip install asn1tools deepdiff
             - run: make test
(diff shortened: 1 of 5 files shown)


changab (approved):
I have no problem with this change. One I thing I would like to add,
That is good to modulize the schema as a sort of extension-based. That would be even better if we can maintain those modules in different files. That also helps the maintenance and easier for review. I tried asn1tools command line to Compile the multiple schema files. I believe the same way is supported in Python module. 
@timsifive I will create a task to modulize schema in the separate files if you agree with the above.
Ah BTW, I can take this work.

PR #42, opened 2021-07-25
Split and organize schema.asn

changab:
Split and organize schema.asn into smaller schema files under /schema for
easy maintenance.
/schema/configuration-structure.asn - Renamed from schema.asn
/schema/debug-extension.asn - Debug extension schema
/schema/hart-extension.asn  - Hart extension schema
/schema/helper-types.asn    - Other generic types schema

rvcs searches the files under /schema as the standard schema files.
"--schema" argument is used for the additional schema other than standard
ones. One of use cases is vendors to generate the binary with their schema alone
with the standard schema files.


asn1tools/rvcs.py
```diff
@@ -106,28 +107,50 @@ def cmd_test(args):
 
         print("%s is %dB in UPER" % (path, len(uper)))
 
-def main():
-    default_schema = os.path.join(
-        os.path.dirname(os.path.realpath(__file__)),
-        "../schema.asn")
+schema_help_String  = 'Specify the addtional ASN.1 schema files to the standard\n' +\
```
timsifive:
```suggestion
schema_help_string  = 'Specify the additional ASN.1 schema files to the standard\n' \
```

timsifive: In python, two strings are automatically concatenated so you don't need the + in these lines.

asn1tools/rvcs.py
```diff
@@ -106,28 +107,50 @@ def cmd_test(args):
...
+    for root, dirs, schemafiles in os.walk(os.path.join (
+        os.path.dirname(os.path.realpath(__file__)), "../schema")):
+        for schemafile in schemafiles:
+            if default_schema != "":
+                default_schema = default_schema + ','
+            default_schema = default_schema + root + '/' + schemafile
+
+    parser = argparse.ArgumentParser(formatter_class=RawTextHelpFormatter)
+    parser.add_argument('--schema', default=default_schema,
+            help=schema_help_String)
```
timsifive, line 161:
```suggestion
            help=schema_help_string)
```

asn1tools/rvcs.py
```diff
@@ -106,28 +107,50 @@ def cmd_test(args):
...
+    default_schema = ''
+    for root, dirs, schemafiles in os.walk(os.path.join (
+        os.path.dirname(os.path.realpath(__file__)), "../schema")):
+        for schemafile in schemafiles:
+            if default_schema != "":
+                default_schema = default_schema + ','
+            default_schema = default_schema + root + '/' + schemafile
+
+    parser = argparse.ArgumentParser(formatter_class=RawTextHelpFormatter)
+    parser.add_argument('--schema', default=default_schema,
```
timsifive: This really should be a list. I think you can do it by adding nargs='*'. Then we don't need to worry about dealing with strings with commas in them everywhere.

timsifive (changes requested):
I really like splitting out the schema like that. It will make maintenance a lot easier.

I'm not sure about --schema only specifying additional schemas. There ought to be some way to ignore the defaults altogether. I think the simplest is to interpret a missing --schema option to use the defaults, and every other option requiring every single .asn to be listed. That's really not so hard for the user. (E.g. `rvcs.py --schema /path/to/schema/* custom.asn`.) We can add a feature later that if a directory name is given to --schema, it automatically finds every asn file under that directory to make this even easier. (Do you need that on Windows? Do modern Windows shells still not do file globbing?)

asn1tools/rvcs.py
changab: I found that if we use "nargs=" and have the white space between the input schema files, the following arguments (such as "convert", source jer file name, and target binary file name) are also count as the arguments to "--schema". Unless we put "-schema" as the last option to rvcs.py. Any suggestions?

changab:
> I really like splitting out the schema like that. It will make maintenance a lot easier.
> 
> I'm not sure about --schema only specifying additional schemas. There ought to be some way to ignore the defaults altogether. I think the simplest is to interpret a missing --schema option to use the defaults, and every other option requiring every single .asn to be listed. That's really not so hard for the user. (E.g. `rvcs.py --schema /path/to/schema/* custom.asn`.)

I think we can specify not just the additional schemas. We can allow users to specify the schemas and reapalce the data types (with the identical notation of the data type) defined in the default schema with the one declared in the schemas specified in "--schema". 
However, we don't allow users to disorder the data type sequence or value type defined in the standard schema, they can just have the vendor-defined data type.

> We can add a feature later that if a directory name is given to --schema, it automatically finds every asn file under that directory to make this even easier. (Do you need that on Windows? Do modern Windows shells still not do file globbing?)

Sure. This would be much easier for users.

timsifive:
> I think we can specify not just the additional schemas. We can allow users to specify the schemas and reapalce the data types (with the identical notation of the data type) defined in the default schema with the one declared in the schemas specified in "--schema".
> However, we don't allow users to disorder the data type sequence or value type defined in the standard schema, they can just have the vendor-defined data type.

That all sounds reasonable, but more complex to implement. If you have time for this, that is great. But I'd prefer to quickly get this PR merged. For now can you make the simpler change which defaults to all the usual .asn files, but requires the user to specify every .asn file if they want to do something custom?

changab:
> That all sounds reasonable, but more complex to implement. If you have time for this, that is great. But I'd prefer to quickly get this PR merged. For now can you make the simpler change which defaults to all the usual .asn files, but requires the user to specify every .asn file if they want to do something custom?

Yes, for now, I will split the schema and use the default .asn files under a new folder schema/.
Users can specify the .asn file that could be a totally new schema or the overwrite schema to the usual one.

I will address the "commas" issues after the above is committed.

changab:
rvcs searches the files under /schema as the standard schema files.
"--schema" argument is used for the additional schema or overwrite the default schema.

I will investigate how to use "nargs" to avoid the comma of the given schema files against "--schema".

asn1tools/rvcs.py
```diff
@@ -106,28 +106,103 @@ def cmd_test(args):
...
+        path_elements = schema.split (os.path.sep)
+        if ".asn" in path_elements[len(path_elements)-1]:
+            schema_dict[path_elements[len(path_elements)-1]] = schema
+    return schema_dict
+
+def default_schema_dict():
+    # Walk through ../schema for all standard ASN.1 schema files.
+    default_schema_list = []
+    for root, dirs, schemafiles in os.walk(os.path.join (
+        os.path.dirname(os.path.realpath(__file__)), ".." + os.path.sep + "schema")):
```
timsifive:
```suggestion
        os.path.dirname(os.path.realpath(__file__)), "..", "schema")):
```

asn1tools/rvcs.py
```diff
@@ -106,28 +106,103 @@ def cmd_test(args):
 
         print("%s is %dB in UPER" % (path, len(uper)))
 
+def schema_path_dict(schemas_list):
+    schema_dict = {}
+    for schema in schemas_list:
+        path_elements = schema.split (os.path.sep)
+        if ".asn" in path_elements[len(path_elements)-1]:
```
timsifive:
```suggestion
        if path_elements[-1].endswith(".asn"):
```

asn1tools/rvcs.py
```diff
@@ -106,28 +106,103 @@ def cmd_test(args):
...
+            schema_dict[path_elements[len(path_elements)-1]] = schema
```
timsifive:
```suggestion
            schema_dict[path_elements[-1]] = schema
```

asn1tools/rvcs.py
timsifive: Why check for the .asn extension? Are we silently ignoring some files the user specified?

asn1tools/rvcs.py
```diff
@@ -106,28 +106,103 @@ def cmd_test(args):
...
+    verified_list = []
+    for index, additional_schema in enumerate(additional_schema_list):
+        # Validate the schema file.
+        schema_file = os.path.normpath(os.path.join (os.path.dirname(os.path.realpath(__file__)), additional_schema))
+        if os.path.exists(schema_file):
+            verified_list.append(schema_file)
+        else:
+            print("Schema file: \"" + schema_file + "\" is not exist.")
+    if len (verified_list) == 0:
+        return {}
```
timsifive: schema_path_dict already returns an empty dict when given an empty list.

asn1tools/rvcs.py
```diff
@@ -106,28 +106,103 @@ def cmd_test(args):
...
+def default_schema_dict():
+    # Walk through ../schema for all standard ASN.1 schema files.
+    default_schema_list = []
+    for root, dirs, schemafiles in os.walk(os.path.join (
+        os.path.dirname(os.path.realpath(__file__)), ".." + os.path.sep + "schema")):
+        for schemafile in schemafiles:
+             default_schema_list.append(root + os.path.sep + schemafile)
+    if len(default_schema_list) == 0:
+        return {}
+    else:
```
timsifive: schema_path_dict already returns an empty dict when given an empty list.

asn1tools/rvcs.py
```diff
@@ -106,28 +106,103 @@ def cmd_test(args):
...
+
+    parser = argparse.ArgumentParser(formatter_class=RawTextHelpFormatter)
+    parser.add_argument('--schema',
+            help=schema_help_String)
 
     subparsers = parser.add_subparsers()
 
     parse_convert = subparsers.add_parser('convert',
-            help='Convert between binary and human-readable formats.')
+            help=convert_help_String)
```
timsifive, line 166: This seems like it's just making the code more complex.

asn1tools/rvcs.py
```diff
@@ -106,28 +106,103 @@ def cmd_test(args):
...
-        "../schema.asn")
-
-    parser = argparse.ArgumentParser()
-    parser.add_argument('--schema', default=default_schema)
+    
+    # Build up the default schema dictionary.
+    try:
+        default_schema = default_schema_dict()
+        if len (default_schema) == 0:
+            print("No defualt schemas found.")
```
timsifive:
```suggestion
            print("No default schemas found.")
```

asn1tools/rvcs.py
```diff
@@ -106,28 +106,103 @@ def cmd_test(args):
...
 
     args = parser.parse_args()
+
+    if args.schema:
+        # Build up the dictionary of schemas assigned against to --schema option.
+        schema_dictionary = {}
+        try:
+            # Generate the dictionary of additional schemas.
+            schema_dictionary = additional_schema_dict(args.schema)
+            if len(schema_dictionary) == 0:
```
timsifive, line 184: This should be an error. The user specified a schema with `--schema` and we didn't find any valid files.

asn1tools/rvcs.py
```diff
@@ -106,28 +106,103 @@ def cmd_test(args):
...
+                schema_dictionary = default_schema
+            else:
+                # Combine schema_dictionary with the default schema.
+                for default_schema_name in default_schema:
+                    if default_schema_name not in schema_dictionary:
+                        schema_dictionary[default_schema_name] = default_schema[default_schema_name]
+        except:
+            print("Fail to build up the dictionary of schemas assigned against to --schema option.")
+            sys.exit ()
```
timsifive:
```suggestion
            return 1
```

asn1tools/rvcs.py
```diff
@@ -106,28 +106,103 @@ def cmd_test(args):
...
-    parser = argparse.ArgumentParser()
-    parser.add_argument('--schema', default=default_schema)
+    
+    # Build up the default schema dictionary.
+    try:
+        default_schema = default_schema_dict()
+        if len (default_schema) == 0:
+            print("No defualt schemas found.")
+            sys.exit ()
+    except:
```
timsifive, line 155: It's bad practice to use `except` without specifying which exceptions you want to catch. The main problem is that this also catches KeyboardInterrupt, which happens when the user hits ^C. Generally in that case you just want the program to die.

We can fix this later, but it's yet another subtle thing about python programming.

asn1tools/rvcs.py
```diff
@@ -106,28 +106,103 @@ def cmd_test(args):
...
+                        schema_dictionary[default_schema_name] = default_schema[default_schema_name]
+        except:
+            print("Fail to build up the dictionary of schemas assigned against to --schema option.")
+            sys.exit ()
+    else:
+        schema_dictionary = default_schema
+
+    # Convert schema dictionary to a list.
+    for schema in schema_dictionary:
+        schema_list.append (schema_dictionary[schema])
```
timsifive, line 200:
```suggestion
    schema_list = schema_dictionary.values()
```

timsifive (changes requested):
Thanks. Things look a lot cleaner with a list instead of dictionary.

I have comments below, but they should all be pretty straightforward.

asn1tools/rvcs.py
changab: yes, good catch.

asn1tools/rvcs.py
changab: good catch

asn1tools/rvcs.py
changab: Not embedding those help strings in code makes the code looks clear IMO. Especially when we add more options in the future.

asn1tools/rvcs.py
changab: yes, let's make it more clear later. Now I just throw an exception when "something" is wrong. :)

asn1tools/rvcs.py
changab: **schema_dictionary.values()** returns a view object which is not reflect on the global schema_list. So the global schema_list is always an empty list when we call into other functions.
I don't actually know how to fix it. I am not the Python expert though.

changab:
@timsifive thanks for those good comments on Python language.
 Update at https://github.com/riscv/configuration-structure/pull/42/commits/837a755bafa07c22ef450a076204054f19aec029

asn1tools/rvcs.py
timsifive: Ah, then you want:
`schema_list = list(schema_dictionary.values())`

timsifive (approved):
I'm not a fan of schema_list being global, but let's get this merged so we can continue with the other PRs. I'm planning to do a cleanup pass at some point.

asn1tools/rvcs.py
changab: > Ah, then you want:
> `schema_list = list(schema_dictionary.values())`

I believe I tried this but that doesn't work for the global schema_list.

changab:
> I'm not a fan of schema_list being global, but let's get this merged so we can continue with the other PRs. I'm planning to do a cleanup pass at some point.

Maybe pass to the function or some other ways. Sure, we can clean it up. thanks

PR #45, opened 2021-07-26
Add Zjpm extension.

timsifive:
Based on email from Martin Maas:
https://lists.riscv.org/g/tech-config/topic/configuration_structure_for/84466618?p=,,,20,0,0,0::recentpostdate%2Fsticky,,,20,2,0,84466618


schema.asn
```diff
@@ -465,3 +467,33 @@ BEGIN
...
+      -- is supported at that privilege level.
+      mSupported BOOLEAN DEFAULT FALSE,
+      sSupported BOOLEAN DEFAULT FALSE,
+      vSupported BOOLEAN DEFAULT FALSE,
+      uSupported BOOLEAN DEFAULT FALSE,
+
+      -- 5/6/7 bits (for RV32/64/128) to indicate how many bits from the top are
+      -- maskable.
+      maskBitcount INTEGER (1..128),
+
```
changab: Is that how many continuous bits are maskable?  that could be 5, 6, or 7 bits? Means 3-bits is enough to indicate how many bits can be masked? Or the maskable bits could be not continuously?

timsifive: That's a good question. I thought Martin meant 5 bits are required on RV32 to encode values from 1 to 32, and so on. I won't merge this until somebody familiar with the spec has taken a look at it.

martinmaas: The definition is that fields maskBitcount..31 (or 63, 127) are maskable for RV32 (or RV64, RV128). This choice means that maskBitcount = 0 is the natural way of encoding that the entire pointer word is maskable (and there is no way to encode that no bits are maskable, which would not be valid according to the spec). Maybe we should call the field "maskableBits" instead of maskBitcount and add this explanation in a comment.

Looks good to me otherwise.

timsifive: How about we call this lowestMaskBit?

allenjbaum:
or MskLSB for short?

schema.asn
martinmaas: SGTM

changab (approved):

changab (approved):

PR #47, opened 2021-09-01
Run pylint on our python code.

timsifive:
With multiple devs come multiple styles. Use pylint to make things a bit
more consistent and clean.

* Made schema_list local instead of global.
* Removed try/except. If people regularly run into python stack traces,
  we can catch specific exceptions to deal with the specific cases they
  run into. Otherwise this just hides what's really going on and makes
  debugging harder.


changab (approved):
The cleanup on rvcs.py looks good to me.

PR #48, opened 2021-09-03
Add virtual memory extension schema and example.

changab:


cento:
looks fine to me

daniellustig:
LGTM, thanks!

schema/virtual-memory-extension.asn
```diff
@@ -0,0 +1,19 @@
+Virtual-Memory-Extension
+DEFINITIONS
+   AUTOMATIC TAGS ::=
+BEGIN
+   EXPORTS VirtMem;
+
+   VirtMem ::= SEQUENCE {
+      -- 1 bit to indicate Svnapot standard extension is supported on Sv39, Sv48
+      -- and Sv57 page table entry.
+      svnapot BOOLEAN,
```
timsifive:
```suggestion
      svnapot BOOLEAN DEFAULT FALSE,
```

This way people can omit this bit in the description if it's not implemented.

The same for the two BOOLEANs below.

schema/virtual-memory-extension.asn
```diff
@@ -0,0 +1,19 @@
...
+
+      -- 1 bit to indicate Svpbmt standard extension is supported on Sv39, Sv48
+      -- and Sv57 page table entry.
+      svpbmt  BOOLEAN,
+
+      -- 1 bit to indicate Svinval Standard Extension is supported.
+      svinval BOOLEAN
+   }
```
timsifive, line 19: We should end this SEQUENCE with `, ...` unless we're absolutely sure there will never be any more information from the virtual memory group.

timsifive (changes requested):

schema/virtual-memory-extension.asn
changab: good point. done in the second commit

schema/virtual-memory-extension.asn
changab: sure, done in the second commit

timsifive (approved):

PR #50, opened 2021-09-16
Elaborate on access method a bit.

timsifive:


riscv-configuration-structure-draft.adoc
```diff
@@ -268,9 +268,37 @@ big immediate benefit of a format that supports comments.
...
 == Access Method
 
-The binary configuration structure is memory-mapped in system memory.
-There is a CSR which contains the physical address where the structure
-starts.
+The binary configuration structure is memory-mapped in system memory.  There is
+a CSR which contains the physical address where the structure starts.
+
+TODO: Get a CSR number assigned.
+
```
changab, line 273: Or use the existing CSR, e.g.  reuse mipmid

riscv-configuration-structure-draft.adoc
```diff
@@ -268,9 +268,37 @@ big immediate benefit of a format that supports comments.
...
+it's running on.
+
+This provides hardware implementers several options for embedding the binary
+structure:
+1. The structure can describe all harts, and be accessible over the memory bus.
+All harts have the same memory map and the same CSR contents.
+2. There might be several structures in the system, and different harts are
+pointed to different structures by having different memory maps.
+3. There might be several structures in the system, and different harts are
+pointed to different structures by having different pointers in their CSR.
```
changab: What is the difference between these two? one is "having different memory maps" and another is "having different pointers in their CSR". The pointer of the memory map is retrieved from CSR, thus I think #2 and #3 use cases are the same. no?

riscv-configuration-structure-draft.adoc
```diff
@@ -268,9 +268,37 @@ big immediate benefit of a format that supports comments.
...
+4. A combination of 2 and 3 above could be used.
+
```
changab, line 290: In order to distinguish which method is provided by the processor package. We should declare a boolean in configuration-structure schema that indicates the CSs are aggregated to one structure and all the same if read from any harts. Or the CS is discrete in each hart. Because the hart 0 is mandatory, the boolean can only be provided by CS in hart0. With this FW knows how to read the complete CS, either from one hart, or run code in each hart to gather all CSs.

riscv-configuration-structure-draft.adoc
```diff
@@ -268,9 +268,37 @@ big immediate benefit of a format that supports comments.
...
+Hardware implementers have a lot of flexibility to handle everything from simple
+fixed systems to complex socketed systems. In each case it's straightforward to
+ensure that each hart can read a configuration structure that describes its own
+capabilities.
+
+How to discover information about each hart in a socketed system is an open
+question. Presumably each chip contains a structure it describes itself. It
+would be nice to have some definition of a motherboard configuration structure,
+which could contain pointers to the configuration structures in each chip.  That
+feels outside the scope of what the RISC-V Foundation defines, however.
```
changab, line 300: Agree with this, the platform may provide the hardware strapping to each socket, so the processor on each socket can determine the memory map address of CS without conflicts.

riscv-configuration-structure-draft.adoc
changab: Another method:
There might be several structures in the system, and different harts are pointed to different structures by having the same pointer in their CSR, however, the memory map pointed by CSR is only visible to the local hart.

riscv-configuration-structure-draft.adoc
timsifive: What is mipmid? I can't find it mentioned in the privspec. There is mimpid but that already has a definition that's not compatible with it being a pointer.

riscv-configuration-structure-draft.adoc
timsifive: I think what you just wrote is what I intended by option 2. Perhaps we don't mean the same thing by "memory map." I'll try to rephrase that.

riscv-configuration-structure-draft.adoc
timsifive: That sounds reasonable, but how would this really be used? Like, either firmware can easily communicate with code running on other harts and this isn't really an issue, or it's impossible and then knowing that there are more harts doesn't help.

riscv-configuration-structure-draft.adoc
timsifive: Thinking of your comment above, we could add an optional list of pointers to other structures to the schema. That way systems with a universal memory map but different CSR pointers per chip can get info about all the components in the system.

riscv-configuration-structure-draft.adoc
changab: FW can send the IPI to another hart to run the code to retrieve CS. That is doable and just depends on how do we implement it in FW.

riscv-configuration-structure-draft.adoc
changab: The method I pointed out is the "different harts are pointed to different structures by having the **same**  memory map address". However, the memory map is only visible to the local hart.
These three methods are actually implementation-specific, we just list some design examples.

riscv-configuration-structure-draft.adoc
changab: Typo, it should be **mimpid**. 
Although it already has the definition, we still can have a specific mechanism to get the memory map address using mimpid. For example, write 55aa55aa to mimpid, then read mimpid that returns the memory map address. Then write another 55aa55aa to close the session. This way we don't have to introduce a new CSR or the exception to probe the existence of that CSR.

riscv-configuration-structure-draft.adoc
changab: Not quite understand this method. Can this method support multiple sockets but only one processor is installed? The second socket may be installed with an identical or different SKU processor later in the next power cycle?

allenjbaum:
I would say that in a multi-socket system, either each socket has it's own
structure, or there is a shared version.
How that works is platform specific; it could be programmed by a service
processor, configured by strap pins, etc.

A multi-socket system could always require that one specific socket be
populated (e.g. hart0) and all other sockets will point to that one.
Or, they could all point to something within their own socket - with the
same address (so it's local only) or a socket specific address.

All you need to know is that by the time the hart tries to read it, the CSR
is pointing to something that can be interpreted.
How that happens is platform specific, just as many other properties are
(e.g. reset vector, meaning of reset status, etc).

changab:
> All you need to know is that by the time the hart tries to read it, the CSR is pointing to something that can be interpreted. How that happens is platform specific, just as many other properties are (e.g. reset vector, meaning of reset status, etc).

Agree with this point. The implementation behind this CSR is out of this TG scope.  We just provide some implementation examples.

changab:
> I would say that in a multi-socket system, either each socket has it's own structure, or there is a shared version. How that works is platform specific; it could be programmed by a service processor, configured by strap pins, etc. A multi-socket system could always require that one specific socket be populated (e.g. hart0) and all other sockets will point to that one. Or, they could all point to something within their own socket - with the same address (so it's local only) or a socket specific address.

Agree that is all platform-specific, FW only faces CSR.
BTW, my expectation of FW implementation is FW would have the common code (with the limited OEMs/Vendor specific code to initial HW for the implementation of that CSR) for reading CS from harts in processors.  FW (at least edk2) implementation of reading CS is expected to have a unified solution for all platforms. At this point,  FW needs to know if the CS is aggregated, shared, or discrete from one hart, harts, or each processor.
However, the information of whether the system is multi-sockets may not be the scope of hart CS.

changab:
Sorry, I closed this PR accidentally.

riscv-configuration-structure-draft.adoc
timsifive: That seems relatively complex. In addition, mimpid has number 0xF13 which means it's in the read-only space of CSRs so should not be writable.

But it's a good point that it would be nice to avoid catching exceptions. Looking through the privileged spec more closely, there's mconfigptr which appears to be already set aside for a binary structure. "mconfigptr must be implemented, but it may be hardwired to zero to indicate the configuration data structure does not exist or that an alternative mechanism must be used to locate it." So that's perfect. I'll update the PR to reference mconfigptr.

riscv-configuration-structure-draft.adoc
timsifive: We're definitely talking about the same thing, we just need to figure out language that makes that clear to everybody. How about:

"There might be several structures in the system. Each hart has the same address in mconfigptr. The memory system provides a different configuration structure at that address depending on which hart is performing the access."

I don't really want to use the word "memory map" because to me it means something like the picture in http://www.sharetechnote.com/html/EmbeddedSystem_MemoryMap.html which is not really relevant here.

riscv-configuration-structure-draft.adoc
timsifive: What does IPI mean in this context? How does the communication between harts occur? Whatever that communication mechanism is, can also be used to check if there are other harts.

riscv-configuration-structure-draft.adoc
timsifive: Assume a motherboard with 2 sockets. Let's say there's this memory map (implemented on the motherboard):
0x8000: structure containing pointers to 0x10000 and 0x11000
0x10000: points to the configuration structure embedded in the chip inserted into socket 0.
0x11000: points to the configuration structure embedded in the chip inserted into socket 1.

Each chip to be inserted into such a system has mconfigptr set to 0x8000, and contains a configuration structure that describes its own harts.

So... firmware runs. It reads mconfigptr, looks at 0x8000, and then looks at 0x10000 and 0x11000, and now has a complete picture of the system. (I'm not sure how to deal with the fact that each chip should have unique hart IDs. You can have chips know which socket they are in and adjust hart ID, but you can't really have them adjust the configuration structure because the structure is too complex.)

riscv-configuration-structure-draft.adoc
changab: yes, this sounds to me better. thanks.

riscv-configuration-structure-draft.adoc
changab: That refers to the interprocessor interrupt, which triggers the software interrupt on other harts. FW can use SBI to send the IPI to other harts for executing specific code. FW also knows how many harts are provided on the system because each hart runs fw code when it boots up.  So there are at least two solutions to read the discrete CS. How does FW read CS from hart 0 or other harts is implementation-specific and out of this TG scope.

riscv-configuration-structure-draft.adoc
changab: Ok, instead of a boolean, a pointer to a list works as well.
But we don't need to have a list entry that points to another socket. The table of lists is only for the local processor package. The processor itself should be socket agnostic. A table of lists that point to CSs of each local hart makes sense to me.

Below is the flow of FW reads the CS, when each hart in each physical processor boots up
FW reads the value from mconfigptr 
   - If it returns 0, then means no there is no CS implemented in this hart.
   - If non zero then FW read the table of lists. It could be a one-entry list or multiple list entries which depends on the implementation. (the size of the table or the number of lists is TBD)
   - FW read the CS from each list.
 FW can then aggregates all CS it reads from harts.

 Above can be applied on below CS implementations,
- One hart describes the entire CS of all harts in the processor package.
- Each hart describes its own CS.
- Or mixup

The address of CS could be,
 - the local (private) address to the hart
 - the global address to each harts.

changab (commented):
Ah, I forget this, Greg or someone else proposes mconfigptr CSR for the configuration structure. But this is a new CSR, right? Or it was already defined somewhere?

riscv-configuration-structure-draft.adoc
timsifive: I'll make a separate PR for adding pointers to the structure.

timsifive: Actually, I already had added it to this PR. See structurePointers in Configuration.

timsifive:
mconfigptr is mentioned in the privileged spec: https://github.com/riscv/riscv-isa-manual/blob/master/src/machine.tex#L2259

allenjbaum:
Ah, a newer version of the spec than I had downloaded, sec 3.1.17; I was
wondering if that was what was going on.
I had gotten the idea that it was a ratified  CSR, but not yet.

changab (approved):

PR #52, opened 2021-10-01
Initial support for scalar crypto v1.0.0-rc3

ben-marshall:
Hi there

This PR adds support to the configuration structure for the scalar cryptography extensions v1.0.0-rc3. The following extensions are added:
- `Zbkb` - Basic bitmanip instructions for cryptography
- `Zbkc` - Carryless multiply
- `Zbkx` - `xperm*` instructions
- `Zknd` - NIST AES decrypt instructions
- `Zkne` - NIST AES encrypt instructions
- `Zknh` - NIST SHA2 hash instructions
- `Zksed` - SM4 encrypt/decrypt instructions
- `Zksh` - SM3 hash function instructions
- `Zkt` - Constant time mode indicator
- `Zkr` - Entropy source.

Each extension is a single bit in the config structure. I've tried structure them one file per extension. Looking forward to a world where RISC-V has _many_ extensions, this seemed the most sensible, rather than hiding multiple extensions in fewer files.

One open question: The `Zkr` extension - access to the entropy source is always possible in M mode, and optionally enabled in S/U mode based on corresponding bits in the `mseccfg` register. These bits can be hardwired or read/writeable. I have not encoded whether these bits are read/write in the config structure. My rationale was that SW only needs to know if the bits exist (because `Zkr` is indicated by the config structure), and can go and test them itself. So encoding them in the config structure would be redundant information. I hope this is the right call.

Please let me know if there's anything which would improve this PR, I'm happy iterating on it.

Thanks,
Ben

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,3 +12,8 @@
 ## Test changes
 
 `make test` or `./asn1tools/rvcs.py test examples/*.jer`
+
+Several python libraries are required to run the tests.
+See the `requirements.txt` file, or run `pip3 install -r requirements.txt`
+
+to install them all automatically.
--- a/examples/example.jer
+++ b/examples/example.jer
@@ -186,6 +186,14 @@
             "fastInt": {
                 "mModeTimeRegAddr": 4660,
                 "mModeTimeCompRegAddr": 4660
+            },
+            "zk" : {
+                "zbkbSupported": true,
+                "zbkxSupported": true,
+                "zkndSupported": true,
+                "zkneSupported": true,
+                "zknhSupported": true,
+                "zkrSupported": true
             }
         }
     ],
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,2 @@
+asn1tools
+deepdiff
--- a/schema/configuration-structure.asn
+++ b/schema/configuration-structure.asn
@@ -7,7 +7,9 @@ BEGIN
       Tuple, Triple, PossibleValue, PhysicalAddress FROM Helper-Types
       Isa, PrivModes, PrivSatps, Privileged FROM Hart-Extension
       VirtMem FROM Virtual-Memory-Extension
-      Zjpm FROM Zjpm-Extension;
+      Zjpm FROM Zjpm-Extension
+      Zk   FROM Zk-Extensions
+      ;
 
    CustomOctetString ::= SEQUENCE {
       -- Ideally we'd build something based on information object classes.
@@ -49,6 +51,7 @@ BEGIN
       custom SEQUENCE OF Custom OPTIONAL,
       virtmem VirtMem OPTIONAL,
       zjpm Zjpm OPTIONAL,
+      zk    Zk  OPTIONAL,
       ...
    }
 
--- a/schema/zk-extensions.asn
+++ b/schema/zk-extensions.asn
@@ -1,0 +1,35 @@
+Zk-Extensions
+DEFINITIONS
+    AUTOMATIC TAGS ::=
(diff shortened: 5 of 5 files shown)


timsifive:
Thanks for getting this going, Ben!

I think all crypto extensions should live in the same type. As it is, your change will add 10 bits to every hart description, even if no crypto extension is implemented. (Well, 20 as written but we could fix that.) If we group them together, then it will take 1 bit if no crypto extension is implemented, and 11 bits if any of them are implemented. (Then making the crypto type extensible adds one more bit.) That is more scalable, especially as we add more and more extensions.

(Aside: I'm still pondering if we should encode things different at top-level, in which case the logic is something like: No overhead if nothing is used, but 8 bits for every option that is used. By grouping that would change to no overhead if nothing is used, and 19 bits if any subset of options is used. Assuming it's common to implement at least 2 of the crypto extensions together, that is almost a wash and at 3 options the grouped version is much more compact.)

ben-marshall:
Hi @timsifive 

So I _think_ I've implemented your recommendations wrt. grouping all of the scalar crypto extensions into a single extensible type / sequence. Experimenting with the example does indeed show this comes out smaller. I'm still getting my head around ASN1 so if there are more compact ways to encode things per hart, then let me know.

In terms of common groups of crypto extensions we expect people to implement, these would likely be the "macro" extensions which group together smaller extensions. Namely:
- `Zk` : `Zkn*, Zkr, Zbk*`.
- `Zks` : `Zks*, Zkr, Zbk*`.

In terms of number of crypto extensions implemented then, it's most commonly going to be zero or at-least three.

I'll have a closer look at the top level structure and see if I have any thoughts about it. One thing to note though is that the cryptography community has an _extremely_ low opinion of ASN.1 because it's been the source of many bugs, both functional and security related. There might be a bit of work to do explaining it's usage as part of the RISC-V config structure.

Cheers,
Ben

README.md
```diff
@@ -12,3 +12,12 @@
 ## Test changes
 
 `make test` or `./asn1tools/rvcs.py test examples/*.jer`
+
+Several python libraries are required to run the tests.
+See the `requirements.txt` file, or run
+
+```
+pip3 install -r requirements.txt
+```
```
timsifive:
```suggestion
See the `requirements.txt` file, or run `pip3 install -r requirements.txt`
```

timsifive (approved):
This looks good to me. I can't think of a way that nicely adds an option for the macro extensions that doesn't make the schema much more complex. I'd be inclined to keep this as-is.

One remaining question: Does it make sense to call this "Crypto" instead of "Zk"? That way any future crypto extensions can extend this configuration block, instead of making a whole new one.

timsifive:
> I'll have a closer look at the top level structure and see if I have any thoughts about it. One thing to note though is that the cryptography community has an extremely low opinion of ASN.1 because it's been the source of many bugs, both functional and security related. There might be a bit of work to do explaining it's usage as part of the RISC-V config structure.

That's an interesting point. Certainly ASN.1 is complex so I can totally see where the bugs come from. Since in this use case software is parsing a structure provided by the hardware (presumably from ROM) I think that reduces the security concern. The argument for ASN.1 is basically that it does exactly what we want (compact binary representation, flexible type system) without us reinventing this particular wheel.

changab:
Looks good to me.

changab (approved):

ben-marshall:
Thanks for the approvals folks! I've implemented Tim's suggested change and squash/rebased. Hopefully this means you can merge it.

PR #55, opened 2021-10-20
Use SEQUENCE OF CHOICE instead of OPTIONAL.

timsifive:
This reduces the size of small descriptions (of presumably small
systems), but increases the size of larger descriptions.

Assuming we're going to end up with 53 or so extensions, descriptions
that use 8 or fewer of them will benefit up to 6 bytes from this change.
Descriptions that use 26 extensions in this example would pay 19 bytes.
(If all extensions are used, the cost of this change is 45 bytes.)

Is that a worthwhile trade-off?

In our specific examples, the numbers change:
examples/debug_module.jer stays the same at 18B
examples/example.jer grows from 93B to 97B

### DIFF
--- a/examples/example.jer
+++ b/examples/example.jer
@@ -9,64 +9,70 @@
                         ]
                     }
                 },
-                "debug": {
-                    "trigger": [
-                        {
-                            "index" : {
-                                "range": [
-                                    {
-                                        "start": 0,
-                                        "length": 3
+                "extensions": [
+                    {
+                        "debug": {
+                            "trigger": [
+                                {
+                                    "index": {
+                                        "range": [
+                                            {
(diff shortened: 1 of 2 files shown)


allenjbaum:
You are talking about tiny numbers of bytes. That might make a difference
in the small systems - where you are saving bytes, and be inconsequential
cost to a larger system implementing all the extensions. That sounds like
the right tradeoff.

changab (approved):

timsifive:
At today's meeting we had 3 people estimate how many extensions there will be in the future, and how many will be used in small and large cores. All 3 estimates would result in this change reducing the structure size for small cores, with relatively little increase for large cores.

PR #58, opened 2021-11-01
Add scaffolding to support signing the structure.

timsifive:


changab:
Let's just wait for TEE's response. Maybe we don't need this change.

timsifive:
Agreed on holding off on this for now. If I do go forward with this change, I should look at https://datatracker.ietf.org/doc/html/rfc3279#section-3 as an example of signatures in ASN.1.

timsifive:
RFC 3279 (through 3281) contain much more stuff than we need to simply record a signature. E.g. stuff about certificates and public keys. I don't think it's appropriate to try to use that schema here.

changab:
Yes, I agree. We should only have the schema for signature and some bitmap that indicate the algorithm to measure the signature.
certificates and the public key is released by the vendor to OEM, and OEM keeps it somewhere as the implementation-specific (BMC or FW storage) for the measuring.

changab (approved):

PR #59, opened 2021-11-05
Limit physical address size to 128 bits.

timsifive:
Saves space in the structure, and surely 128 bits is enough for any
address.


pdonahue-ventana:
640KB is enough for anyone. ;-)

changab (approved):

PR #62, opened 2021-11-18
Move Use Cases higher up.

timsifive:
Use Cases is more part of the rationale/introduction part of the
document, and should come before we talk about implentation details.


changab (approved):
looks good

PR #66, opened 2021-12-07
Add Socket type.

timsifive:
Explicitly represent sockets, and the translation required to
incorporate the embedded Configuration Structure into the main one.


changab:
I am confused about this PR. How a configuration structure in a physical processor or HART can describe the configuration of other sockets. CS should single processor information.  The sockets configuration should be the platform (board) stuff, right?

Other comments in the source code.

schema/configuration-structure.asn
```diff
@@ -83,13 +83,33 @@ BEGIN
       -- Encoded as (extended:1'h1 index:5'h0 length:... contents:...)
    }
 
+   Socket ::= SEQUENCE {
+      -- Represents a socket in a system, where different things might be
+      -- plugged in which can contain their own Configuration Structure.
+      -- (This could also be used for other IP baked into the design.)
```
changab: The hart/core CS to describe the socket on the platform seems doesn't make sense. CS goes with the processor and it is unpredictable which platform the processor is plugged in.

schema/configuration-structure.asn
```diff
@@ -83,13 +83,33 @@ BEGIN
...
+      -- Hardware in the socket might have memory-mapped devices. This is how
+      -- their addresses are translated by the host board to prevent conflicts
+      -- in the memory map.
+      addressTranslation Translation OPTIONAL
+   }
+
    Configuration ::= SEQUENCE {
       harts SEQUENCE OF Hart OPTIONAL,
       debugModule SEQUENCE OF DebugModule OPTIONAL,
+      sockets SEQUENCE OF Socket OPTIONAL,
```
changab: I think I need more information to understand the reason to have hart/core CS to describe the platform sockets.

schema/configuration-structure.asn
```diff
@@ -83,13 +83,33 @@ BEGIN
...
+
+      -- Hardware in the socket might think it contains harts 0 and 1, but
+      -- depending on which socket they are plugged into, it could be 2 and 3.
+      -- It could even be 79 and 92 in a very strange system.
+      hartIdTranslation Translation OPTIONAL,
+
+      -- Hardware in the socket might have memory-mapped devices. This is how
+      -- their addresses are translated by the host board to prevent conflicts
+      -- in the memory map.
+      addressTranslation Translation OPTIONAL
```
changab: How does FW know that it has to do the translation when the processor is plugged into the specific socket?
I thought this translation could be done by core using HW strapping without FW intervening unless we don't want the core to dynamically change CS.

schema/helper-types.asn
```diff
@@ -84,6 +84,15 @@ BEGIN
       ...
    }
 
+   Translation ::= SEQUENCE OF SEQUENCE {
```
changab, line 87: Let's clarity the rationale of this PR first then get back to review this schema.

timsifive:
I'm going to rework this so it's more clear.

The fundamental problem is that the configuration structure cannot be easily modified by a chip that goes in a socket, because e.g. hartid in the structure is a variable-length field. So we need to give the parser some information it can use to get a complete view of the system based around a static configuration structure on the chip that goes in a socket.

changab:
Ok, thanks.
My thought is that system-based information can be provided by the DT. Otherwise, the processor may not be compliant with two different boards. e.g two-socket and four-socket boards.
Another thing comes to my mind is do we have to translate hart ID in the physical processor?  hart ID suppose is the logical number to the physical processor, right?

timsifive:
> My thought is that system-based information can be provided by the DT. Otherwise, the processor may not be compliant with two different boards. e.g two-socket and four-socket boards.

In the scheme I'm trying to describe a processor can be compliant with two-socket and four-socket boards. The processor just needs to be able to find the ancestor CS, and work its way down from there.

> Another thing comes to my mind is do we have to translate hart ID in the physical processor? hart ID suppose is the logical number to the physical processor, right?

I'm not sure what exactly you mean. Each processor must somehow come up with a hart ID that is unique in the system. I assume this works by checking some pins which tell the processor which socket it's placed in. Once the processor has that hart ID, the CS needs to describe that processor referring to the correct hart ID.

riscv-configuration-structure-draft.adoc
```diff
@@ -153,6 +153,80 @@ schema of Configuration Structure. It describes, in varying levels of detail, th
...
+can describe about each component.) Each component also has access to a
+system-wide bus, and implements some memory-mapped peripherals. The mappings
+shown are into the system-wide address space, although it happens in stages.
+I.e. each daughterboard maps its second socket to 0x2_0000 locally, and the 
+motherboard maps the second daughterboard to 0x2_00000, which results in CPU
+Socket 3 having a global address of 0x22_0000.
+
+The pointers in the various Configuration Structure look as follows:
+
+* motherboard: CS @0x3000, childPointers=0x10_3000, 0x20_3000
```
changab: Is the motherboard CS store at 0x3000 somewhere on the motherboard but not in the processor??

riscv-configuration-structure-draft.adoc
```diff
@@ -153,6 +153,80 @@ schema of Configuration Structure. It describes, in varying levels of detail, th
...
+*** vector processor 1: CS @0x3000, ancestorPointer=0x3000
+
+If you combine this with the mappings listed previously, the system bus address of each CS is:
+
+* motherboard: 0x3000
+* server module: 0x10_3000
+* server processor: 0x11_3000
+* vector module: 0x20_3000
+* vector processor 0: 0x21_3000
+* vector processor 1: 0x21_3000
```
changab: what is the value in mconfigptr for this case?  The server processor 0x11_3000 is also the value reported in mconfigptr? and Vector processor 0's mconfigptr is 0x21_3000?
Does this mechanism also use mconfigptr to point to the CS reported by the processor?

changab: ok, I am trying to imagine how the firmware works with this mechanism,
Is that FW reads **mconfigptr** to harvest the core/hart CS. In the CS, there is an **ancestorPointer** which indicates the ancestor CS is @0x3000 (which is the motherboard CS)?
In the ancestor CS, a **childPointers** may exist which contains **hartIdTranslation** that describes the way to translate the hart ID on this socket?

riscv-configuration-structure-draft.adoc
timsifive: Yes. Each of these modules has its own CS.

riscv-configuration-structure-draft.adoc
timsifive: mconfigptr could just be 0x3000. Then it ends up pointing to the top-level CS, and maybe ancestorPointer is unnecessary.

The algorithm you describe is what I have in mind.

changab:
Thanks, Tim The diagram makes this proposal clear, now I understand what's the mechanism for the socket-based system.
I have a couple of questions,
First of all, I thought the socket-based processor system refers to the sockets that are able to plug in the processor directly to the motherboard.
Here we have the intermediary which is called daughterboard that supports two socket-based processors. The daughterboard is a kind of expander board attached to a socket on the motherboard with two processors installed.
Instead of having four sockets on the motherboard to support the 4-P system, two daughterboards support a maximum of 2x2 processors.  Maybe the idea is some other devices may also be implemented on the daughterboard?
I think two multi-processor solutions should be accepted on the platform, 
1. Motherboard directly attached socket-based processor
2. Daughterboard-based multiple-socket processor
I think the childPointer CS proposed here can support both scenarios. However, we should describe it in the spec as the childPointer use cases.

changab:
Because there is no hardware method for accessing CS on the daughterboard and motherboard, that makes sense to define 0x3000 (or some other values) in the platform spec to access childPointers that is described in motherboard CS and daughterboard CS.
Firmware goes through the child pointers in motherboard CS and access to the child CS. However, how does FW know the daughterboard CS exists? The childPoitner may contain all 0xff, all zeros, or other garbage values if the daughterboard or the processor is not installed. We have to define a signature at the beginning of CS or some other ways that indicate a valid CS exists, right?

changab:
Another question is, there may have conflicts of the translation if two identical processors are plugged into either daughterboard or motherboard because they have the same translation CS for HART ID or device memory map address. Is this concern valid?

Processor somehow needs the capability to dynamically update the translation values in CS according to the hardware trapping in order to avoid conflicts no matter how many processors are installed on the system. 
For example, the hardware strapping may indicate the Multiplier of each socket, thus the processor can determine the value of "multiple" and update the value in Translation CS.
We don't want to see the value in CS that can be updated dynamically especially when the CS is signed. However, the main CS for the processor HW features and the Translation CS could be located in different regions. Translation CS can be updated accordingly without impacting the main CS. Furthermore, the main CS still can have childPointer points to Translation CS (0x3000).

timsifive:
I intentionally went with a more complex example with an intermediary board, to make sure that the scheme can support it. If it can support this, it will also support a simpler design where processors are directly plugged into a motherboard. I also believe the scheme scales to more intermediaries, if that is ever necessary. A nice side effect of that is that the same scheme can be used when components are integrated from multiple vendors.

Good point about the header. We currently have effectively a constant value for the first byte. I had thought that it was 0, but experimentally is not. We need to revisit that. Do you think one byte that's not 0 or 0xff is good enough?

I thought of a reason to keep ancestorPointer. It's helpful if there is a system where 0x3000 isn't always where the parent config is. Maybe that is true for the processors, but in the motherboard it's at 0x5000 or something. In that case the daughterboard can make sure that there is a CS at 0x3000 (because it knows its children expect that) with an ancestorPointer of 0x5000 (because it knows its parent expects that).

> Another question is, there may have conflicts of the translation if two identical processors are plugged into either daughterboard or motherboard because they have the same translation CS for HART ID or device memory map address. Is this concern valid?

The translation is specified in the parent CS, not the child. I'll see if I can update the figure with that info. It's getting kind of crowded in there.

changab:
> Good point about the header. We currently have effectively a constant value for the first byte. I had thought that it was 0, but experimentally is not. We need to revisit that. Do you think one byte that's not 0 or 0xff is good enough?

I would prefer to have something like ASCII "_CS_" in the header If there is no size concern. Otherwise, one-byte pattern 0x5a or two-byte pattern 0x55aa would be fine.

changab:
> The translation is specified in the parent CS, not the child. I'll see if I can update the figure with that info. It's getting kind of crowded in there.

ok, I get it wrong.
If so, two processors would have the same HART ID translation in the daughterboard use case,  as the HART ID translation table is described in both daughterboard's CS which is the parent of the processor, right?  Or the parent of the daughterboard has the second-level translation?
It would not be a problem to the motherboard directly attached processor use case. However, I think it still has the problem with the translation table in two or more identical components that are installed in the system, even though the component is the parent of other components. Right?

changab:
Another question is do we expect the hardware configuration should be changed according to the translation table?
For example, is the value in mhartid CSR be replaced with the new value calculated according to the translation table?
That is fine if mhartid still keeps the same because the translated value is used by FW to distinguish the HART in each physical processor (for example send the IPI to a specific HART in a specific processor). But if we take  CLINT/ACLINT as an example, the hardware configuration should be changed for the MP systems.
Assume there are 2-P systems and each processor has the address 0x200000 that maps to CLINT msip CSR for IPI and mtimecmp CSR for the timer interrupt. Even though there is a translation table to relocate the second CLINT to 0x400000, the processor has to take a new value and decode the CLINT memory I/O cycle at a different address. I think CLIC has the similar problem, the base address of CLIC memory map indicated in mclicbase CSR should be changed as well on the MP system. Is this a valid concern?

changab:
> I'll see if I can update the figure with that info. It's getting kind of crowded in there.

I can help to redraw the diagrams to sync up the style with the current one in spec.

timsifive:
> I would prefer to have something like ASCII "CS" in the header If there is no size concern. Otherwise, one-byte pattern 0x5a or two-byte pattern 0x55aa would be fine.

I just ran a quick experiment. If I run an ASN.1 decoder on all zeros, you get an empty structure: `[  "configuration",  {} ]` If you run an ASN.1 decoder on all ones you quickly get a parse error: `Top: Bad length determinant fragmentation value 0xff.` Those both seem like reasonable outcomes, without having to write any extra code to check a special header. So I think we can simply say the memory must be all 0 or all 0xff, which seem like a very reasonable limitation.

> If so, two processors would have the same HART ID translation in the daughterboard use case, as the HART ID translation table is described in both daughterboard's CS which is the parent of the processor, right? Or the parent of the daughterboard has the second-level translation?

Translation is cumulative going through the levels. I'll add hartid translation to the diagram. E.g. when you walk from the Motherboard CS to the 8-hart vector CS in Daughterboard 1, Socket 1, you first add 32 to each hartid you see, and then add 16. So that processor's effective hartid will be 48--55.

> Another question is do we expect the hardware configuration should be changed according to the translation table?

No. The CS describes the hardware

> For example, is the value in mhartid CSR be replaced with the new value calculated according to the translation table?

No. The system cannot work that way. hartids must be unique from the very start, and the hardware has to ensure that using some other mechanism. (I suppose the hardware could ensure that only one hart can run code from power-up, and that hart parses the CSs and updates mhartid in each hart, but that's not part of what we're specifying.)

> Assume there are 2-P systems and each processor has the address 0x200000 that maps to CLINT msip CSR for IPI and mtimecmp CSR for the timer interrupt. Even though there is a translation table to relocate the second CLINT to 0x400000, the processor has to take a new value and decode the CLINT memory I/O cycle at a different address.

I don't want to solve that problem. How is that addressed in existing systems? I suspect the socket spec has some pins that communicate to the processor which socket they are in, and they use that to adjust addresses, but I might be wrong.

changab:
> Those both seem like reasonable outcomes, without having to write any extra code to check a special header. So I think we can simply say the memory must be all 0 or all 0xff, which seem like a very reasonable limitation.

Ok, I am fine with this. Can we clearly define how many bytes of 0x00 or 0xff is sufficient to tell the parse there is no CS?
(Update: ignore this, I just saw that in your new commit)

changab:
> Translation is cumulative going through the levels. I'll add hartid translation to the diagram. E.g. when you walk from the Motherboard CS to the 8-hart vector CS in Daughterboard 1, Socket 1, you first add 32 to each hartid you see, and then add 16. So that processor's effective hartid will be 48--55.

Clear now. thanks

changab:
> > Another question is do we expect the hardware configuration should be changed according to the translation table?
> 
> No. The CS describes the hardware
> 
> > For example, is the value in mhartid CSR be replaced with the new value calculated according to the translation table?
> 
> No. The system cannot work that way. hartids must be unique from the very start, and the hardware has to ensure that using some other mechanism. (I suppose the hardware could ensure that only one hart can run code from power-up, and that hart parses the CSs and updates mhartid in each hart, but that's not part of what we're specifying.)

Yes, this is what do I mean. Hardware must somehow update the HW configuration according to some mechanism. I originally thought HW can also refer to CS and update itself accordingly, but I think no vendor would like to embed the ASN.1 parser in the processor. So hardware strapping would be necessary.

> 
> > Assume there are 2-P systems and each processor has the address 0x200000 that maps to CLINT msip CSR for IPI and mtimecmp CSR for the timer interrupt. Even though there is a translation table to relocate the second CLINT to 0x400000, the processor has to take a new value and decode the CLINT memory I/O cycle at a different address.
> 
> I don't want to solve that problem. How is that addressed in existing systems?

Hardware strapping is a common way, mainly used for the processor but not usually used on devices.

> I suspect the socket spec has some pins that communicate to the processor which socket they are in, and they use that to adjust addresses, but I might be wrong.

So there are two key points,
1. CS in the component mainly gives FW an idea of how to access hardware
2. Hardware strapping, the hardware itself requires the capability to update the HW configuration physically.

The above two must be synced, thus FW/SW can access hardware without problems.
Although I still can see the challenge of #2 (how many hardware strappings processor has to probe? HardID, CLINT, more? and on both mainboard and daughterboard), we can move forward and come out the first version of spec for the socket-based platform.
We don't have to solve the HW configuration updates in CS spec, but we need to mention this in the spec because this is very important to the entire system for FW to access hardware.

Some actions items.
1. Define 0x3000 as the address to the component CS in the corresponding spec, e.g privilege spec and platform spec.
2. Seems we have to define the component type for *addressTranslation*, thus FW knows the translation is for which device?

timsifive:
> We don't have to solve the HW configuration updates in CS spec, but we need to mention this in the spec because this is very important to the entire system for FW to access hardware.

What do you want to mention? The CS should describe the hardware. That's the whole point. I don't know what we could write that doesn't simply repeat that.

> Some actions items.
> 
> Define 0x3000 as the address to the component CS in the corresponding spec, e.g privilege spec and platform spec.

Maybe. It would be cool to have this in the platform spec, but I suspect it's too late to pick an address that will work for every RISC-V system.

> Seems we have to define the component type for addressTranslation, thus FW knows the translation is for which device?

Why? The pointer is simply to another CS, which describes the device. The motherboard shouldn't need to know what kind of device is plugged into it. (At least not as far as CS is concerned. There may be other limitations imposed by the hardware.)

riscv-configuration-structure-draft.adoc
```diff
@@ -155,6 +155,89 @@ It describes, in varying levels of detail, the
...
+*** vector processor 0: CS @0x3000, ancestorPointer=0x3000
+*** vector processor 1: CS @0x3000, ancestorPointer=0x3000
+
+If you combine this with the mappings listed previously, the system bus address of each CS is:
+
+* motherboard: 0x3000
+* server module: 0x10_3000
+* server processor: 0x11_3000
+* vector module: 0x20_3000
+* vector processor 0: 0x21_3000
```
changab: In below, is vector processor 1 system bus address 0x22_3000?

changab:
> Why? The pointer is simply to another CS, which describes the device. The motherboard shouldn't need to know what kind of device is plugged into it. (At least not as far as CS is concerned. There may be other limitations imposed by the hardware.)

Is that any memory map devices (e.g. ACLINT, UART) implemented in the child component are translated (or rebased) using addressTranslation declared in the parent CS? If so, then my question is not a problem.

changab:
> What do you want to mention? The CS should describe the hardware. That's the whole point. I don't know what we could write that doesn't simply repeat that.

I think it would be something like "The hardware configuration of the socket-based components should be adjusted recursively according to the translation declared in the parent CS"

But one thing I would like to confirm,
In the example that follows the socket-based mechanism, the system-wide memory space of 0x10_0000-0x1f_ffff is decoded by Socket 0 and 0x20_0000-0x2f_ffff is decoded by Socket 1 on the motherboard.
So, does daughterboard 0 decodes 0x1_0000-0x1_ffff for its socket 0 and 2_0000-0x2_ffff for its socket 1, but not 0x11_0000 and 0x12_0000?  In the end, the Single-hart server processor decodes 0x3000 for its CS but it does not decode the system-wide memory address 0x11_3000?
That says, for example, the base address of ACLINT in the processor is still the local address but not the translated address? although the FW uses the translated address to access it.

timsifive:
> Is that any memory map devices (e.g. ACLINT, UART) implemented in the child component are translated (or rebased) using addressTranslation declared in the parent CS? If so, then my question is not a problem.

Any address mentioned in a child CS is translated according to the ChildStructure.addressTranslation. But... that leads to a problem. What if there are memory-mapped peripherals local to a module, not exposed on a system bus? I think that would need a bit on every address, whether it is to be translated or not (defaulting to true).

> That says, for example, the base address of ACLINT in the processor is still the local address but not the translated address? although the FW uses the translated address to access it.

Yes, I think. So maybe there's ACLINT at 0x4000 in a processor, which will end up at 0x11_4000, 0x12_4000, 0x21_4000, or 0x22_4000 depending on which socket the processor is plugged into. The processor would access it through the global address.

changab:
> Any address mentioned in a child CS is translated according to the ChildStructure.addressTranslation. But... that leads to a problem. What if there are memory-mapped peripherals local to a module, not exposed on a system bus? I think that would need a bit on every address, whether it is to be translated or not (defaulting to true).

The memory-mapped peripheral local to a module still can be accessed through the system bus according to the bridge-like mechanism we propose here, right.  All local addresses include the memory-mapped local to a module are translated using the parent CS addressTranslation. The drawback I can think of is the code that runs on another processor can still access the memory-mapped peripheral local to a module on this processor. Unless that module local memory-mapped space is restricted to expose to the system bus for some reason. Do you have an example of memory-mapped peripherals local to a module?

changab:
> Yes, I think. So maybe there's ACLINT at 0x4000 in a processor, which will end up at 0x11_4000, 0x12_4000, 0x21_4000, or 0x22_4000 depending on which socket the processor is plugged into. The processor would access it through the global address.

Make sense now.

timsifive:
> The memory-mapped peripheral local to a module still can be accessed through the system bus according to the bridge-like mechanism we propose here, right.

Maybe. We're not proposing something. We're imagining systems that people might build, and making sure that CS would be able to describe them.

> All local addresses include the memory-mapped local to a module are translated using the parent CS addressTranslation. The drawback I can think of is the code that runs on another processor can still access the memory-mapped peripheral local to a module on this processor. Unless that module local memory-mapped space is restricted to expose to the system bus for some reason.

These are all issues for the system designer to consider. We just want to be able to describe the system, whatever reasonable decisions they make.

> Do you have an example of memory-mapped peripherals local to a module?

* Server daughterboard: Ethernet
* Vector daughterboard: fast memory
* any module: temperature sensor

changab:
> These are all issues for the system designer to consider. We just want to be able to describe the system, whatever reasonable decisions they make.

This should be spoken out somewhere in spec then the system vendor/OEM knows how to design the hardware along with the socket type CS. Maybe a new spec for socket-based platform or platform spec.
I think I am good at understanding socket-type CS. I will go through the last content you committed before next week CS meeting.

changab (approved):
Approve for the first round review.

changab:
> Maybe. We're not proposing something. We're imagining systems that people might build, and making sure that CS would be
> able to describe them.

I understand that we are not proposing the hardware solution. But we should have the hardware implementation example as the reference so the stuff we add to CS spec for the socketed-type platform looks more clear. We can remove the example section once the hardware design for supporting socketed-type is clearly defined in another spec.

Issue #68, opened 2022-01-07
Convert the project to docs-template


jjscheel:
The Config documentation need to be in the RISC-V format.  I've discussed getting there with @timsifive and he agreed to let me take a look at this.

Four things were needed:
1. Import of docs-template project into a subdirectory (docs-template)
2. Addition of symbolic links in the resources/ and resources/images/ directories for the fonts/ themes/ and images/ (RISC-V)
3. Updates to the Makefile to use more options on asciidoctor build
4. Updates to the main .adoc file to include more variables for the build

I got it working and inadvertently pushed my updates back to main in commit #[19430d9](https://github.com/riscv/configuration-structure/commit/19430d9e4ae7e2ebe82e15c8f1adcc7e0aca228a).  Sorry about that!!!

If anyone sees problems with the updates, feel free to reach out or add information to this issue.  I'll leave it open until we've had a chance to discuss it more and prove things are working as expected.

Again, my apologies for the push to master!
-Jeff

jjscheel:
FYI, per item 1 above, the import was done using Git Submodules.  Note that this essentially "links" to that repo at a commit value.  Over time, we will want to update this.  More information on Git Submodules can be found [here](https://git-scm.com/book/en/v2/Git-Tools-Submodules).

As a result of this change, the next update people make locally will require two additional commands (once).  They are:

```
git submodule init
git submodule update
```

If one does not do this after updating, the build will fail with an error like:
```
$ make
asciidoctor-pdf -v \
-r asciidoctor-diagram \
-a toc \
-a compress \
-a pdf-style=resources/riscv-themes/riscv-pdf.yml \
-a pdf-fontsdir=resources/riscv-fonts \
-o riscv-configuration-structure-draft.pdf riscv-configuration-structure-draft.adoc
asciidoctor: ERROR: could not locate or load the built-in pdf theme `resources/riscv-themes/riscv-pdf.yml'; reverting to default theme
No such file or directory - notoserif-regular-subset.ttf not found in resources/riscv-fonts
Use --trace for backtrace
make: *** [Makefile:13: riscv-configuration-structure-draft.pdf] Error 1 
```

timsifive:
I've reverted your change because I couldn't easily get the new stuff to build on my machine. I got some horrendous error regarding asciidoctor-diagram (after I installed it).

jjscheel:
[...]

PR #70, opened 2022-01-07
Convert the project to the RISC-V docs-template.

jjscheel:
Merging updates to adopt the RISC-V document template (#68)

Signed-off-by: Jeff Scheel <jeff@riscv.org>


timsifive:
Can you add the `git submodule` explanation you put into #68 to the README.md?

cetola:
@jjscheel If you remove the "-r asciidoctor-diagram" flag from the Makefile edits in this commit, the PR will build just fine. 

This brings up the larger question. Do you think it's better to keep the builds consistent across repos, or allow repos to simplify their builds if applicable? This repo is a good example of this idea.

I don't have a strong opinion but if we allow folks to keep their builds as simple as possible it does save us GitHub action cycles (which if free, are limited for our Org). The tradeoff is that each spec's build script will be unique rather than mirroring the docs-spec-template.

jjscheel:
@cetola, my personal preference is to only enable the features/functions one needs.  So, your suggestion might be better than re-working the Git Actions.

First thing tomorrow, I'll update the Makefile and either submit a new PR or try and add it to this one.

Sound, ok, @timsifive?

changab:
Not sure if org has this or not? Could we have a label for PR that triggers Git action to convert the asciidoc to PDF, with the repo-specific asciidoc attributes and build options? Then merge the PDF to repo automatically. Thus we don't have to set up the local environment for asciidoc-pdf and run it locally.

jjscheel:
Will cleanup and re-submit.

jjscheel:
@changab, FWIW, I think your question is a fair one.  The build actions were added by @timsifive very recently (just before my submission), not by my proposed patch.  As such, I'd suggest we move your discussion to a new issue.  

In addition, I'd note that the [docs-dev-guide](https://github.com/riscv/docs-dev-guide) project does what you suggest.

PR #72, opened 2022-01-11
Address some of Greg's feedback.

timsifive:
Also slight reorganisation to (hopefully) make more sense.

### DIFF
--- a/riscv-configuration-structure-draft.adoc
+++ b/riscv-configuration-structure-draft.adoc
@@ -72,7 +72,7 @@ other features determine what kind of operations the user can perform
 discoverable, although many require writing a value and checking the
 result to see whether support exists.
 
-Any structure that’s accessible from M-mode software will already be
+Any structure that's accessible from M-mode software will already be
 accessible by the debugger. There might be a structure embedded in the
 Debug Module itself which is only accessible by the debugger.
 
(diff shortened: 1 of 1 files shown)


allenjbaum:
Sorry to have to pile on here. I think you're too close to this, so you're not explaining things that are obvious to you.
This doc says:
        Part of the socket standard is the fact that the root CS exists at 0x3000. 
Which socket standard is this? IS this a proposed platform spec standard?
This doc doesn't actually say that mconfigptr is a CSR - that should be clear.

From this example I think that there is 
 - some kind of ROM on the motherboard (at address 0x3000) that points back to each daughtercard, and
 - another on each daughtercard that points to a ROM on each socket,  and that contains the CS for that socket 
   (it could be inside the socket, or external). 
At least, that's how I interpret it (while acknowledging my superpower of finding a different explanation for whatever someone describes).
Since I would expect that all sockets could be identical, that means that - at the socket, the address decoding would be 
ignoring bits 21:20 (assuming 2 daughter cards), and bits 17:16 when routing to the ROM.

Alternately, there could be a ROM on the motherboard that contains all of the CSes, and it ignores all irrelevant address bits to keep it small.

But, in the former case: what happens when someone tries to follow the breadcrumbs to some socket, and that socket isn't even plugged in? (a bus error?). 
In the latter case: the ROM is accessible - but the socket isn't populated. How is that discovered?

changab (approved):

timsifive:
I'll address Allen's comments, which don't really pertain to this PR, later somewhere else.

PR #84, opened 2022-01-13
Document the process to extend the schema.

timsifive:
I just made this process up. Consider is a proposal.

### DIFF
--- a/riscv-configuration-structure-draft.adoc
+++ b/riscv-configuration-structure-draft.adoc
@@ -154,6 +154,29 @@ https://github.com/riscv/configuration-structure/tree/master/schema. The
 top-level type is Top, defined in
 https://github.com/riscv/configuration-structure/blob/master/schema/configuration-structure.asn[configuration-structure.asn].
 
+=== Extending the Schema
+
+The schema is part of this specification, but might be updated without requiring
+the Configuration Structure Specification needs to be ratified again. The
+process is as follows:
+
+1. A task group decides they need to extend the schema to incorporate additional
+design decisions.
+2. The task group prepares a pull request (PR) against
+https://github.com/riscv/configuration-structure that updates the schema.
+3. The Configuration Structure Task Group reviews this pull request and works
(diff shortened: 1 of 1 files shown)


riscv-configuration-structure-draft.adoc
```diff
@@ -154,6 +154,29 @@ https://github.com/riscv/configuration-structure/tree/master/schema. The
...
+Backward compatible means that new parsers will continue to work with existing
+Configuration Structures.
+Forward compatible means that old parsers will skip over data described by the
+new schema improvement.
+4. The task group treats this PR as part of their specification, and it will go
+through whatever process is required to get this PR incorporated into the
+specification. Once that process is complete, they update the PR indicating so.
+5. The CS Task Group merges the PR. It is now part of the official schema, and
+can be used immediately.
+
```
changab, line 179: No problem with the content however this seems the scope of Readme but not the spec.

timsifive: I can see how it's really a process *about* the document, so doesn't belong in the document itself. However, this procedure needs to be documented and ratified. Currently there is a procedure for getting specifications ratified, and a github README is not part of that procedure. So it seems easier to include this process in the spec itself.

If we put it in the README, how would we get that ratified?

changab: Good question. Maybe we can have this in temporary and tell Tech-chair we would like to move this paragraph to README because that is about the process but not the spec. They may give us some recommendations.

changab (approved):

allenjbaum:
The "philosophical questions should be directed to jeff scheel and
stephano; they can probably figure out where it out to go.
We had something similar for the ACTs, and it went into a policy document,

riscv-configuration-structure-draft.adoc
timsifive: Jeff Scheel prefers to keep the process in the spec.

PR #90, opened 2022-01-18
Add Overview section.

timsifive:
Also edited the Access Method section, where some of the Overview stuff
came from.


changab (approved):
I sent the updated figure to your mailbox for the 1P  configuration structure use case.
Could you please replace the figure in this section with the new one?
Just keep the multi-P figure under /figures, we can add the complex example later.

PR #93, opened 2022-01-26
Move Implemenation Examples into the example section.

timsifive:
Also expand on each use case a bit.

### DIFF
--- a/riscv-configuration-structure-draft.adoc
+++ b/riscv-configuration-structure-draft.adoc
@@ -296,7 +296,7 @@ it's running on.
 
 There could be a single system-wide CS, or a more complex arrangement for either
 single core or multicore systems, using the ChildStructure type discussed in
-<<_configuration_structure_pointers>>.
+<<Configuration Structure Pointers>>.
 
 == Contents of Configuration Structure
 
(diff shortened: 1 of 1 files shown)


riscv-configuration-structure-draft.adoc
```diff
@@ -403,32 +403,47 @@ and then later that same element in a different type describing that same
...
+
+Different harts might be pointed at different CSs by giving them unique
+mconfigptr values. Alternatively, the memory system could provide a different CS
+depending on which hart is accessing a given address. There are many
+possibilities
+
+=== Tree of Structures
+
+When combining IP from multiple vendors it might be helpful to use structure
+pointers (see <<_configuration_structure_pointers>>) instead of combining all
```
changab: Please check if this link works or not. It seems problematic here.

timsifive: Hmm... it works in vscode adoc mode, and also in the PDF. But github doesn't seem to recognize this. I'll see if I can change it so it works on github also.

timsifive: Fixed. :-)

changab (approved):

PR #105, opened 2022-03-29
Make rvcs C library using asn1c

timsifive:
The rough build process is:
1. asn1c generates .c and .h files from the schema.
2. We compile those files, along with rvcs.c into a .a archive.
3. During the automated build, the result is combined into a tar file that contains all of that together. You can look at that by clicking the Actions tab, then "build C library," then pick a recent build, and finally download it from Artifacts.

I don't want the generated .c files in the tar file but I haven't figured out yet how to make that happen.

The resulting decoder should be correct, but it is large (59015b text, 29600b data, 115b bss). Part of that is that it includes ASCII names for each types, support for various asn.1 formats we don't need, and in general generates functions for each type instead of doing something more efficient. Those are all limitations in asn1c.


lib/README.md
```diff
@@ -0,0 +1,8 @@
+# Configuration Structure C Library
```
irmafmz, line 1: If you could please add the workflow example of how to use the output C library with an specific Configuration Structure that would be great

lib/README.md
```diff
@@ -0,0 +1,8 @@
...
+
```
irmafmz, line 2: It would be great if you  could please add the svg diagram you showed us today, many thanks in advance

irmafmz (approved):
No further comments, thanks for addressing all of our questions

PR #107, opened 2022-10-23
IFM:Adding SEQUENCEs for the standard unprivileged and privileged ext…

irmafmz:
…ensions, as well as for the non-standard extensions.


allenjbaum (commented):
How do you deal with base ISA's that can change, e.g. an RV64I that can change XLEN to 32 and become an RV32I?  Someone may want to take advantage of that to run RV32I code on an RV64I hart.
Ditto for RV32E (that is a weird one, since it's basically RV32I with registers X16..X31, which means it is perfectly legal for them to behave exactly the same as RV32I)
I think this means that a;; of the baseISA choices can be supported at the same time, so it cna't br a choice.

schema/configuration-structure.asn
```diff
@@ -34,25 +34,114 @@ BEGIN
...
-      -- this way.
-      isa Isa OPTIONAL,
+      }, -- baseISA
+      baseIsaVersion Version,
+
+      stdUnprivExtensions SEQUENCE {
+         m M OPTIONAL, -- M: Integer Multiplication and Division
+         a A OPTIONAL, -- A: Atomic Memory Operations
+         f F OPTIONAL, -- F: Single-Precision Floating-Point.  Implies Zicsr
+         d D OPTIONAL, -- D: Double-Precision Floating-Point.  Implies F
```
timsifive: The way this SEQUENCE gets encoded, is that there will be one bit for every OPTIONAL type, and if that bit is one then the contents of the type will follow. That means that right now this type will take 52 bits to encoded just to say none of the. For types that added after `...` in the future, the cost is more than one bit per option. The minimum size of an encoded structure is directly proportional to the number of extensions there. That number will only continue to go up. Instead I prefer the SEQUENCE OF CHOICE construction used previously. It has more overhead per type that is present, but near 0 overhead for types that are not present.

https://lists.riscv.org/g/tech-config/files/2021%20Meetings/2021-10-11%20Configuration%20Structure%20Task%20Group%20Slides.pdf has a slide showing real numbers for this trade-off.

schema/configuration-structure.asn
```diff
@@ -34,25 +34,114 @@ BEGIN
...
+         zknh   Zknh   OPTIONAL,   -- Zknh: NIST Suite: SHA2 Hashing.
+         zkr    Zkr    OPTIONAL,   -- Zkr: Entropy Source Extension
+         zks    Zks    OPTIONAL,   -- Zks: ShangMi Algorithm Cryptography.  Implies Zbkb, Zbkc, Zbkx, Zksed, and Zksh
+         zksed  Zksed  OPTIONAL,   -- Zksed: ShangMi Suite: SM4 Block Cipher.
+         zksh   Zksh   OPTIONAL,   -- Zksh: ShangMi Suite: SM3 Hashing.
+         zkt    Zkt    OPTIONAL,   -- Zkt: Data Independent Execution Latency
+         ...
+      } OPTIONAL, -- stdUnprivExtensions
+
+      stdSupLevelExtensions SEQUENCE {
```
timsifive: I understand the logical reason for putting this in its own section, but I think it makes more sense to roll this into the structure above, and just use a clear block of comments to make the separation. Extensions are already all in the same namespace so there is no chance that you'll have conflicts that require them to be separate. Doing this will also keep the structure smaller if you switch to SEQUENCE OF CHOICE as opposed to have multiples of them.

If there will only ever be a handful of these top-level categorizations then it's fine the way you have it, but I assume things will only ever grow.

schema/configuration-structure.asn
```diff
@@ -34,25 +34,114 @@ BEGIN
...
+         sscofpmf Sscofpmf OPTIONAL, -- Sscofpmf: Count Overflow and Mode-Based Filtering
+         sstc   Sstc   OPTIONAL,   -- Sstc: Supervisor-mode Timer Interrupts
+         sv32   Sv32   OPTIONAL,   -- Sv32: Page-based Virtual Memory, 32-bit
+         sv39   Sv39   OPTIONAL,   -- Sv39: Page-based Virtual Memory, 39-bit
+         sv48   Sv48   OPTIONAL,   -- Sv48: Page-based Virtual Memory, 48-bit
+         sv57   Sv57   OPTIONAL,   -- Sv57: Page-based Virtual Memory, 57-bit
+         svinval Svinval OPTIONAL, -- Svinval: Fine-Grained Address-Translation Cache Invalidation
+         svnapot Svnapot OPTIONAL, -- Svnapot: NAPOT Translation Contiguity
+         svpbmt Svpbmt OPTIONAL,   -- Svpbmt: Page-Based Memory Types
+      } OPTIONAL, -- stdSupLevelExtensions
```
timsifive:
```suggestion
         svpbmt Svpbmt OPTIONAL,   -- Svpbmt: Page-Based Memory Types
         ...
      } OPTIONAL, -- stdSupLevelExtensions
```

schema/configuration-structure.asn
```diff
@@ -34,25 +34,114 @@ BEGIN
...
+
+      stdHypLevelExtensions SEQUENCE {
+         h H OPTIONAL, -- H: Hypervisor
+      } OPTIONAL, -- stdHypLevelExtensions
```
timsifive:
```suggestion
         h H OPTIONAL, -- H: Hypervisor
         ...
      } OPTIONAL, -- stdHypLevelExtensions
```

schema/configuration-structure.asn
```diff
@@ -34,25 +34,114 @@ BEGIN
...
+
+      stdMachLevelExtensions SEQUENCE {
+      } OPTIONAL, -- stdMachLevelExtensions
```
timsifive:
```suggestion
      stdMachLevelExtensions SEQUENCE {
         ...
      } OPTIONAL, -- stdMachLevelExtensions
```

schema/configuration-structure.asn
```diff
@@ -34,25 +34,114 @@ BEGIN
...
+
+      nonStdExtensionsExist Boolean,
```
timsifive: You don't need this because nonStdExtensions is already OPTIONAL.

schema/configuration-structure.asn
```diff
@@ -34,25 +34,114 @@ BEGIN
...
+      nonStdExtensions SEQUENCE OF {
+         nonStdExtension PrintableString, -- Prefixed with a single “X”, ordered alphabetically and separated from other multi-letter extension names by an underscore on the RISC-V ISA name strings 
+         nonStdExtensionVersion Version OPTIONAL
+      } OPTIONAL, -- nonStdExtensions
```
timsifive: This doesn't include a feature for adding extra information for the non-standard extension. The Custom type tries to do the same thing you did here, for vendor extensions. It includes some black box data for the custom feature to use, if they want.

Almost certainly it's better to define a type (including its name) for non-standard extensions, and then here to make a SEQUENCE OF that type. Putting all the non-standard extensions together into a single string seems unnecessarily complex to deal with for software when structured data is readily available.

timsifive (commented):
I have some ASN.1 feedback, but nothing on the list of extensions itself.

schema/configuration-structure.asn
irmafmz: Hi @timsifive , thank you for the link with the meeting notes of 2021-10-11, I have changed the extension sequences to SEQUENCE OF CHOICE

schema/configuration-structure.asn
irmafmz: Sounds good  @timsifive ;  I have deleted nonStdExtensionsExist from configuration-structure.asn

schema/configuration-structure.asn
irmafmz: Hi  @timsifive , please see the  image below for our  conversation tomorrow, many thanks
image

irmafmz:
Hi @timsifive , here's the link to Mark's document "Profiles & Bases & Extensions".  The changes that I showed today reflect the worksheet  "Profiles", columns "extension only exists if the extension in this column exists" and "implies not".  Many thanks in advance for your Review Comments.
Mark's document "Profiles & Bases & Extensions":
[https://docs.google.com/spreadsheets/d/1A40dfm0nnn2-tgKIhdi3UYQ1GBr8iRiV2edFowvgp7E](https://docs.google.com/spreadsheets/d/1A40dfm0nnn2-tgKIhdi3UYQ1GBr8iRiV2edFowvgp7E)

schema/configuration-structure.asn
```diff
@@ -34,25 +34,161 @@ BEGIN
       }
    }
 
+   Version ::= SEQUENCE {
+      major	INTEGER ( 0 | 1 | ... ),
```
timsifive: Is there any reason to define major like this, instead of just an INTEGER? It allows for a more compact encoding when it's 0 or 1, I suppose.

schema/configuration-structure.asn
```diff
@@ -34,25 +34,161 @@ BEGIN
...
+      minor	INTEGER
```
timsifive: IIRC, INTEGER is 4 bytes by default. Can we make these `INTEGER RANGE (0..255)`? Surely nobody will ever need more than 255 versions. ;-)

schema/configuration-structure.asn
```diff
@@ -34,25 +34,161 @@ BEGIN
...
 
-      -- Extensions that will not be common on small platforms should go here.
-      -- When these extensions are not used they take no space at all.
-      extensions SEQUENCE OF CHOICE {
-         debug Debug,
-         custom SEQUENCE OF Custom,
-         zjpm Zjpm,
-         zk Zk,
+      -- Base Integer ISA indicating the supported address space size in bits
+      baseIsa CHOICE {
```
timsifive: I think you said you were going to make this a SEQUENCE OF CHOICE.

schema/configuration-structure.asn
```diff
@@ -34,25 +34,161 @@ BEGIN
...
+         ...
+      }, -- baseISA
+      baseIsaVersion Version,
+
+      stdUnprivExtensions SEQUENCE OF CHOICE {
+         multiplication CHOICE {
+            m     M     OPTIONAL, -- M: Integer Multiplication and Division
+            zmmul Zmmul OPTIONAL, -- Zmmul: Multiplication Subset of the M extension
+            ...
+         } OPTIONAL, -- multiplication
```
timsifive: I have mixed feelings about this.

On the one hand, it's good to use the structure so that it is impossible to specify invalid configurations.

On the other hand, it can be annoying for someone writing a description that most extensions are just listed in stdUnprivExtensions and others are in their own space like multiplication. The only reason they're in their own space is to enforce certain kinds of constraints, which is nice-to-have but not a requirement for our structure.

I'm also unsure what we can do in the future if the constraints among extensions change. Imagine a future Zmfuture extension, which can coexist with Zmmul but not with M. I think your only option is to put it outside of this multiplication CHOICE so that both Zmfuture and Zmmul can be selected. But then the organisation is even more convoluted, because some multiplication extensions are in the multiplication CHOICE, while others are not.

My preference is to not try to enforce dependency/conflict requirements using the structure here, keeping all the extensions in a single flat SEQUENCE OF CHOICE. A separate tool can be created to check that a given description does not violate any requirements. Such a tool can contain arbitrarily complex rules.

schema/configuration-structure.asn
```diff
@@ -34,25 +34,161 @@ BEGIN
...
+            d D,          -- D: Floating-Point, Double-Precision.  Implies F
+            zdinx  Zdinx, -- Zdinx: Floating-Point, Double-Precision in Integer Registers X.  Implies Zfinx
+            ...
+         } OPTIONAL, -- fpDoublePrecision
+         fpSinglePrecision CHOICE {
+            f F OPTIONAL, -- F: Floating-Point, Single-Precision.  Implies Zicsr
+            zfinx  Zfinx, -- Zfinx: Floating-Point, Single-Precision in Integer Registers X.
+            ...
+         } OPTIONAL, -- fpSinglePrecision
+         q Q OPTIONAL, -- Q: Quad-Precision Floating-Point.  Implies D
```
timsifive: I don't think you ever need a value inside a CHOICE to be OPTIONAL. If the thing isn't implement, simply don't choose it at all.

schema/configuration-structure.asn
```diff
@@ -34,25 +34,161 @@ BEGIN
...
+               svinval   Svinval   OPTIONAL, -- Svinval: Fine-Grained Address-Translation Cache Invalidation
+               ...
+            } OPTIONAL, -- ss1p12Dependent
+            ...
+         } OPTIONAL, -- ss1p12Dependent
+
+         ...
+      } OPTIONAL, -- stdPrivExtensions
+
+      nonStdExtensions SEQUENCE OF CHOICE {
```
timsifive:
```suggestion
      nonStdExtensions SEQUENCE OF SEQUENCE {
```

schema/configuration-structure.asn
```diff
@@ -34,25 +34,161 @@ BEGIN
...
+         nonStdExtension PrintableString, -- Prefixed with a single “X”, ordered alphabetically and separated from other multi-letter extension names by an underscore on the RISC-V ISA name strings
```
timsifive: I thought this field is for a single extension. But the comment talks about separating extensions with an underscore.
Which is it?

irmafmz:
> How do you deal with base ISA's that can change, e.g. an RV64I that can change XLEN to 32 and become an RV32I? Someone may want to take advantage of that to run RV32I code on an RV64I hart. Ditto for RV32E (that is a weird one, since it's basically RV32I with registers X16..X31, which means it is perfectly legal for them to behave exactly the same as RV32I) I think this means that a;; of the baseISA choices can be supported at the same time, so it cna't br a choice.

Hi @allenjbaum , thank you for your comment.  The base ISA has been  changed to be a SEQUENCE OF CHOICE to accommodate for the case you mention above, thank you

schema/configuration-structure.asn
irmafmz: As discussed at the meeting of Nov/7, the  stdSupLevelExtensions, stdHypervisorLevelExtensions and stdMachineLevelExtensions have been rolled into a single stdPrivExtensions SEQUENCE OF CHOICE.  Thanks

schema/configuration-structure.asn
irmafmz: Thanks for the feedback @timsifive .  I have deleted nonStdExtensions, and added name and version fields to the Custom SEQUENCE.

schema/configuration-structure.asn
irmafmz: See the following comment below, thanks

schema/configuration-structure.asn
irmafmz: Done

schema/configuration-structure.asn
irmafmz: Hi @timsifive , yes, I did.  And, yes, it's done, thanks for the reminder

schema/configuration-structure.asn
irmafmz: Right, the reason it was there it was because before our meeting of Nov/7, stdUnprivExtensions was a SEQUENCE, not a SEQUENCE OF CHOICE.  I have deleted all OPTIONAL clauses from all CHOICE structures.  Thanks

schema/configuration-structure.asn
irmafmz: nonStdExtensions was deleted, and name and version fields where added to the Custom SEQUENCE.  For details please see the following comment, thank you
https://github.com/riscv/configuration-structure/pull/107#discussion_r1020952318

schema/configuration-structure.asn
irmafmz: Yes, thank you for the Review Comment @timsifive .  nonStdExtension is for a single extension, yet the comment is talking about more than one extension.  I have modified the comment to not talk about multiple extensions.  Good catch

schema/configuration-structure.asn
irmafmz: Hi @timsifive , the extensions are now a plain list in a SEQUENCE OF CHOICE syntax, thanks for your input

timsifive: This multiplication construct (as well as fpDoublePrecision, etc.) hasn't been changed in the PR.

timsifive:
At the last meeting I think you mentioned you defined all the various extension types, but I don't see them in this PR yet. That will be necessary for all the checks to pass.

irmafmz:
> At the last meeting I think you mentioned you defined all the various extension types, but I don't see them in this PR yet. That will be necessary for all the checks to pass.

Hi @timsifive , please check again, I have just pushed up the next set of changes.  Many thanks in advance for your Review Comments, Irma

schema/zk-extensions.asn
```diff
@@ -1,35 +0,0 @@
-Zk-Extensions
```
timsifive, line 1: I've created a zk-extensions branch that we can use to easily find this file again, if we want it in the future for some reason.

schema/configuration-structure.asn
```diff
@@ -32,27 +34,110 @@ BEGIN
          customOctetString SEQUENCE OF CustomOctetString,
          ...
       }
+      ...
```
timsifive:
```suggestion
      },
      ...
```

It's probably worth your time to get `make test` to work on your system so you don't end up pushing files with syntax errors.

schema/configuration-structure.asn
```diff
@@ -32,27 +34,110 @@ BEGIN
...
-         zjpm Zjpm,
-         zk Zk,
+      -- Base Integer ISA indicating the supported address space size in bits
+      baseIsa SEQUENCE OF CHOICE {
+         rv32I Rv32I,
+         rv64I Rv64I,
+         ...
+      }, -- baseISA
+
+      stdUnprivExtensions SEQUENCE OF CHOICE {
```
timsifive, line 133: This looks right, and much cleaner than before. Thanks.

schema/std-priv-extensions.asn
```diff
@@ -0,0 +1,299 @@
...
+        version Version OPTIONAL,
+        ...
+    }
+
+END -- H-Extension
+
+Shcounterenw-Extension -- Shcounterenw
+DEFINITIONS
+    AUTOMATIC TAGS ::=
+BEGIN
```
timsifive, line 18: Is it even allowed to have multiple DEFINITIONS in a single file like that? How does the IMPORTS statement know which file to look for in this case?

I'm a little on the fence as to whether it makes sense to put all stdpriv extensions into a single file. Having one file per extension feels a little excessive, since most of them will be place holders. But all in one will lead to an ever-growing file. Ideally we can split these up into functional areas, or something. I'm not sure where to draw the line. This is likely a case where no solution is perfect.

schema/std-unpriv-extensions.asn
```diff
@@ -0,0 +1,715 @@
...
+
+Zkt-Extension -- Zkt: Data Independent Execution Latency
+DEFINITIONS
+    AUTOMATIC TAGS ::=
+BEGIN
+    EXPORTS Zkt;
+
+    Zkt ::= SEQUENCE {
+        version Version OPTIONAL,
+        ...
```
timsifive, line 766: Just want to confirm that every single one of these extensions looks exactly like this one, and I didn't miss an odd one out somewhere.

That is a good starting point for future work where individual extensions might need to add additional optional configuration data.

timsifive (changes requested):
This looks like good progress.

There's a conflict between this branch and `master`. Can you resolve that, either by rebasing this one onto `master` or by merging `master` into this one?

schema/std-unpriv-extensions.asn
irmafmz: yes

schema/std-priv-extensions.asn
timsifive: The IMPORTS statement doesn't look in files. It looks for modules. The tools find all the .asn files and parse them all. Then the IMPORTS statement refers to the modules defined in those files.

timsifive:
This should fix the remaining test problems:
```
diff --git a/examples/example.yaml b/examples/example.yaml
index 9a7eaf4..ffa67bb 100644
--- a/examples/example.yaml
+++ b/examples/example.yaml
@@ -146,16 +146,16 @@ configuration:
   - hartId: { single: [ 0, 2, 4 ] }
     baseIsa:
       - rv32I:
-          - version: { major: 1, minor:  0 }
-        rv64I:
-          - version: { major: 1, minor:  0 }
+          version: { major: 1, minor:  0 }
+      - rv64I:
+          version: { major: 1, minor:  0 }
   - hartId: { single: [ 1, 3 ] }
     baseIsa:
       - rv64I:
-          - version: { major: 1, minor:  5 }
+          version: { major: 1, minor:  5 }
   - baseIsa:
     - rv64I: {}
-    zk: {}
+    #zk: {}
         # zbkbSupported: true
         # zbkcSupported: false
         # zbkxSupported: true
```

irmafmz:
Done
```
irma.t.flores-mendozaibm.com@Irmas-MBP configuration-structure % git diff
diff --git a/examples/example.yaml b/examples/example.yaml
index 9a7eaf4..ffa67bb 100644
--- a/examples/example.yaml
+++ b/examples/example.yaml
@@ -146,16 +146,16 @@ configuration:
   - hartId: { single: [ 0, 2, 4 ] }
     baseIsa:
       - rv32I:
-          - version: { major: 1, minor:  0 }
-        rv64I:
-          - version: { major: 1, minor:  0 }
+          version: { major: 1, minor:  0 }
+      - rv64I:
+          version: { major: 1, minor:  0 }
   - hartId: { single: [ 1, 3 ] }
     baseIsa:
       - rv64I:
-          - version: { major: 1, minor:  5 }
+          version: { major: 1, minor:  5 }
   - baseIsa:
     - rv64I: {}
-    zk: {}
+    #zk: {}
         # zbkbSupported: true
         # zbkcSupported: false
         # zbkxSupported: true
irma.t.flores-mendozaibm.com@Irmas-MBP
```

schema/configuration-structure.asn
```diff
@@ -21,6 +103,9 @@ BEGIN
    }
 
    Custom ::= SEQUENCE {
+      customExtName PrintableString, -- Prefixed with a single “X”
+      customExtVersion Version OPTIONAL,
```
timsifive:
```suggestion
      name PrintableString, -- Prefixed with a single “X”
      version Version OPTIONAL,
```

It's already implied that it's for a custom extension because of the type that it is in.

timsifive: Also, it seems unnecessary to prefix the name with X. It's just wasting typing/space. We're already in the custom extension section.

timsifive (changes requested):

timsifive (approved):

PR #108, opened 2023-01-11
Pylint fixes.

timsifive:
New version of pylint means new things to clean up. No functional change.


timsifive:
Included in #109, so we don't have to force anything past failing checks.

PR #110, opened 2023-01-30
Add example of custom data.

timsifive:
And fix up the tests to not fall down when binary data is present.


timsifive:
@irmafmz can you review this?

irmafmz (approved):

PR #111, opened 2023-03-20
Address feedback by Allen Baum

timsifive:


allenjbaum:
Hmm; the Zbb example isn't quite right either. There is no guarantee that
an unimplemented instruction will trap.
Determining whether PMP registers are implemented, and how many - that
works (e.g.
 - if       a read of  PMPCFG3 in mmode traps, then  PMP is not
implemented.
- elseif a read of PMPCFG4 in mmode  traps, then only 16 PMP CSRs are
implemented, - else  64 are implemented
(and then you get to test each one to see if its permanently disabled or
not, and exactly which options it supports. while not cutting yourself off
at the knees....)

timsifive:
What are the possible behaviors of executing `andn` then?
1. Zbb is implemented and it does AND with inverted operand.
2. The instruction traps.
3. Something else?

That seems like it only makes it harder to detect whether Zbb is implemented.

timsifive:
I just skimmed the PMP part of the spec, and that seems almost as hard to detect as hardware triggers. So that's a better example in any case.

allenjbaum:
Yes indeed, any of those 3 actions can occur.
#2 and #3 indicate that Zbb is not implemented (though #3 may not actually
be usable;
for all you know it is a custom op that jumped off into the weeds and kills
your boot process.)
 #1 tells you nothing unfortunately; maybe they've only implemented some of
the easy ops, and not all? There is no way of telling.

allenjbaum:
PMP is easy in comparison to Zbb. The results are unequivocal. IT either
traps or it doesn't.
If it doesn't' then it is implemented (though it doesn't tell you
everything, just that it exists
- it is possible that all the CSRs are readonly-zero,which isn't terribly
useful).
There are more details that may be very difficult to test, though (e.g. any
range limitations on the address field, primarily,
and whether the lock field is functional - which you can test for, but is
nearly impossible to recover from without a reset.

timsifive:
Yikes. All this underlines just how badly we need a configuration structure. Thanks for the explanation.

So you do you like the PR as it is now?

allenjbaum:
I got completely lost on this:
Fields can be defined the same way that types are defined, or can refer to
a custom type.
We generally prefer to use a custom type instead of using nested `SEQUENCE`
or `CHOICE` constructs.

First, I don't think "generally prefer" belongs in a spec, except as
non-normative, and with an explanation why
Second: I need an example; I can't see where this would apply.
Third: this mentions "custom" types - but there is no previous mention of
them, or any explanation of how a custom type gets defined, or
nor why a custom type is necessarily better than a sequence or choice.

timsifive:
> First, I don't think "generally prefer" belongs in a spec, except as non-normative, and with an explanation why

I'm just removing this line. It doesn't seem worth a bunch of confusing language that basically says "don't nest too deeply". This is really just a style thing, which most people will pick up when copying examples, and otherwise reviewers can point out and trivially resolve.

> Third: this mentions "custom" types - but there is no previous mention of them, or any explanation of how a custom type gets defined

Custom types are the ones initially referred to. Added that.

> , or nor why a custom type is necessarily better than a sequence or choice.

In this section I don't think we should go into "why better." It's just giving a brief overview of syntax.

allenjbaum:
This looks fine now. I'm not a reviewer, so can't do much more than comment here.

timsifive:
Thanks, Allen.

PR #124, opened 2024-09-11
Wmat patch 1

wmat:
This change refactors the .adoc files and adds the standard Makefile and GH Action to build the spec.  If you run the GH Action, you get two specs, one in PDF & one in HTML.


sqzsq (approved):
approved

PR #125, opened 2024-10-11
Some minor changes (typos, stylistic, etc)

andreiw:
Reword some of the text.

One thing of note is that the charter referred to a "bootloader" generating DT based on UD data, whereas in a normal non-embedded environment it would be firmware (eg  coreboot, UEFI, w/e), not the OS loader (and uboot is really more firmware than a pure loader, esp. when it is used to modern OSes)


sqzsq (approved):
Approved

sqzsq (commented):
Approved

PR #129, opened 2024-10-11
Update mapping-to-asn1.adoc

andreiw:
Minor edits

(Removed "aforementioned" from DiscoveryAddtionalData as it hasn't been in fact mentioned anywhere in this section)


sqzsq (approved):
Approved

PR #132, opened 2024-10-11
Update schema.adoc

andreiw:
Minor. Really, the "Headers" ought to be split apart further but it's unclear if some of these are ASN.1 generic or specific to UD.

### DIFF
--- a/src/schema.adoc
+++ b/src/schema.adoc
@@ -1,9 +1,11 @@
 == ASN.1 Schema
 
-The Unified Discovery ASN.1 Schema can be broken down in to the following pieces for 
+The Unified Discovery ASN.1 Schema can be broken down into several pieces for 
 better understanding.
 
-The header that declare the data structure:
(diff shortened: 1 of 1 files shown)


sqzsq (approved):
Approved

allenjbaum:
There are numerous typos and minor grammatical errors in the latest;  none
are particularly serious.

I'm not sure the best way to point them out to you, e.g.

DiscoveryAdditonalData

should be

DiscoveryAdditionalData


imaginative hart

should be

imaginary hart


list of extension <-- unless ext3nsion is a type and should be in mono font

should be ?

list of extensions


Is UnifiedDiscoveryEnvelop  deliberately spelled that way, or should it be
UnifiedDiscoveryEnvelope?

Is it a goal to produce a complete (initial) list of the defined/ratified
extensions, parameters, and options for this?
If not... whose responsibility is it to do this?

